PR #58, opened 2014-10-21
Fixed Python version check to keep working for Python 4+

svisser:
This changes ensures the library keeps working in Python 4+.


nicolaiarocci:
thanks for you contirbution. As there's no Python 4 on the horizon, and it won't be for *quite* a while, we'll hold on this one.

svisser:
That's fine, though it's fair to say that the string / integer changes won't be reverted in Python 4.

PR #69, opened 2015-02-18
use str.format instead of % substition so we can replace the error codes

brettatoms:
Using str.format doesn't complain when you try to do a substitution where a doesn't placeholder exists.  This allows me to set errors.ERROR_BAD_TYPE to another string without breaking anything.


brettatoms:
I didn't test this very well.  I'll recreate the PR later.

PR #70, opened 2015-02-18
use str.format instead of & substitution

brettatoms:
Using str.format doesn't complain when you try to do a substitution where a doesn't placeholder exists. This allows me to set errors.ERROR_BAD_TYPE to another string without breaking anything.


nicolaiarocci:
Thanks for this. Unfortunately, I belive that in Python 2.6 `str.format` requires that you number the placeholders, and that's why CI is failing on the Py26 build.

Also, can you take out the `coerce` validator and eventually submit a distinct PR for that one?

brettatoms:
I didn't mean to include the type coercion stuff in here.  I was waiting until I worked out the kinks.  I'll split them out and recreate the PRs.

PR #79, opened 2015-04-17
add the defined data type

goneri:
This data type can be used to validate any defined value.


nicolaiarocci:
This is seems very narrowed down to PostgreSQL/JSON support, I don't think it qualifies for a general purpose validator such as Cerberus core. Seems more suited for a custom validator, one that inherits from Cerberus?

goneri:
Hi Nicola.

Thanks for the review. I will go this way.

nicolaiarocci:
:+1:

PR #81, opened 2015-04-26
Adds Validator-initialization to example with `allow_unknown`

funkyfuture:
This is supposed to clarify that an empty dict has to be provided.
 Also fixes some spellings.


nicolaiarocci:
Thanks!

PR #82, opened 2015-04-26
Allows to validate against a set of types.

funkyfuture:
See added documentation and test for usage. If this doesn't clarify, these should be extended.

- fixes also some minor formatting issues of docstrings
- omit regex-checks on non-string-types


nicolaiarocci:
Thanks this has been merged in https://github.com/nicolaiarocci/cerberus/commit/883942751b0254b2c274c35d300df61f7da57d02

PR #85, opened 2015-05-05
Follow-up of checking against multiple types

funkyfuture:
- improves failure message when testing against multiple types
- ignores `keyschema` when not a mapping
- ignores `schema` when not a sequence
- adds a failing test concerning validating dict and list as allowed value
  in conjunction with keyschema and schema
- some minor code fixes

sadly there's still an issue that is documented as a test. shall i add this documentation?
since i actually don't need this, i will not fix it. unless the universe stops expanding, maybe i find the time then.

also, i seem to miss a point about rebasing. so you should just cherry-pick 875e8f6

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,13 +15,14 @@ Patches and Suggestions
 - Denis Carriere
 - Eelke Hermens
 - Florian Rathgeber
+- Frank Sachsenheim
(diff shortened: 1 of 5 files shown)


funkyfuture:
i will solve this manually.

PR #86, opened 2015-05-05
Follow-up of checking against multiple types

funkyfuture:
- improves failure message when testing against multiple types
- ignores `keyschema` when not a mapping
- ignores `schema` when not a sequence
- adds a failing test concerning validating dict and list as allowed value
  in conjunction with keyschema and schema
- some minor code fixes

sadly there's still an issue that is documented as a test. shall i add this documentation?
since i actually don't need this, i will not fix it. unless the universe stops expanding, maybe i find the time then.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -302,6 +302,15 @@ class Validator(object):
                         if not hasattr(self, '_validate_type_' + value):
                             raise SchemaError(
                                 errors.ERROR_UNKNOWN_TYPE % value)
+                    if 'dict' in values and 'list' in values:
+                        if 'keyschema' in constraints and \
+                            'schema' not in constraints:  # noqa
+                                raise SchemaError('You must provide a compleme'
+                                                  'ntary `schema`')
+                        if 'schema' in constraints and \
+                            'keyschema' not in constraints:  # noqa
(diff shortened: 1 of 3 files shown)


funkyfuture:
why did you close this pr?

nicolaiarocci:
Ups sorry about that!

nicolaiarocci:
> sadly there's still an issue that is documented as a test. shall i add this documentation?

Yes please do update the docs so maybe we spare a couple tickets in the near future :)

funkyfuture:
i guess a solution may be:

- when checking `schema` / `keyschema`
  - if there is a type defined that includes both `dict` and `list`
    - expect `schema` *and* `keyschema` to be defined (so that would belong do schema-validation)
    - ignore `schema` if a `dict` was detected

i think the only barrier for now is the point where `value` shadows an outer loop, so that the `schema`-check can't know which `type`-constraint is its sibling. but that's easy to fix.

disclaimer: i'm ill, my head holds quiet some slime and i didn't look at the code to verify my assumptions. so my lines may contain bs.

funkyfuture:
okay, i fixed the remaining issues, also that `dict`/`list`- & `keyschema`/`schema`-issue. Travis' all green.

PR #87, opened 2015-05-11
Ensures that additional **kwargs of a subclass persist through validation

funkyfuture:
since through a validation-process several child-validators are created, without this patch, these child-validators lose 'knowledge' they may have as a subclass of `Validator`.


funkyfuture:
fixed all glitches. all Travis checks succeeded.

nicolaiarocci:
Thanks this has been merged in https://github.com/nicolaiarocci/cerberus/commit/8d92c756d5026ef3831342d56d00f8548275a893

PR #94, opened 2015-05-23
Fix in case 'type' allows more than two types

funkyfuture:
there was an error in my implementation, leading to false negatives.

on that occasion:
- small code-cleanups
- output the errors in case an `assertSuccess` fails

### DIFF
--- a/cerberus/tests/tests.py
+++ b/cerberus/tests/tests.py
@@ -831,3 +831,37 @@ class TestDockerCompose(TestBase):
 
         document = {'environment': ['VARIABLE=FOO']}
         self.assertSuccess(document, schema)
+
+    def test_one_of_dict_list_string(self):
+        ptrn_domain = '[a-z0-9-]+(\.[a-z0-9-]+)+'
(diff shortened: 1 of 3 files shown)


cerberus/tests/tests.py
```diff
@@ -831,3 +831,37 @@ def test_environment(self):
...
+
+        document = {'image': 'busybox',
+                    'extra_hosts': 'www.domain.net:127.0.0.1'}
+        self.assertSuccess(document, schema)
+
+        document = {'image': 'busybox',
+                    'extra_hosts': ['www.domain.net:127.0.0.1']}
+        self.assertSuccess(document, schema)
+
+        document = {'image': 'busybox',
```
nicolaiarocci: Isn't this redundant (see above test)?

funkyfuture:
thanks for the spot, it was even missing a test-case.

funkyfuture:
pushed the fix.

nicolaiarocci:
thanks!

PR #98, opened 2015-05-24
Adds 'propertyschema'-validation

funkyfuture:
this isn't finished at all yet.

but i'm irritated that the failure-test in `test_a_dict_with_propertyschema` passes, as there is currently nothing implemented that detects an error.

@nicolaiarocci any idea what my mistake is here?

PR #99, opened 2015-05-25
Fixes non-failing tests

funkyfuture:
i came to the conclusion why [a test didn't fail](https://github.com/nicolaiarocci/cerberus/pull/98#issue-80170445).

as any test that doesn't match a requirement fails, the required field is removed from TestBase.schema.
also a test checks the expected behaviour of TestBase.

includes code formattings that are enforced by tighter flake8-rules.

### DIFF
--- a/cerberus/__init__.py
+++ b/cerberus/__init__.py
@@ -8,10 +8,9 @@
 
 """
 
-__version__ = "0.8.2-dev"
-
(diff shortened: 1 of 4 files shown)


nicolaiarocci:
Merged with https://github.com/nicolaiarocci/cerberus/commit/b4b2fb49e0fba60eeeca7ca0501e8500d4cd4a8c

Issue #95, opened 2015-05-23
Proposal: option to purge unknown fields on validation


misja:
I have the need to apart from validating the data to also remove unknown fields. I couldn't see an easy way in _validate to get a direct reference to the document element being parsed so for now I subclass Validator and do a clean post validation based on the errors dict, see [this gist] (https://gist.github.com/misja/d9b88ea849b9ab86af95). I was wondering, has this been discussed before,  would it be something to include, and how to best handle it via _validate?

funkyfuture:
i would add a `del self.document[field]` as alternate action instead of [throwing an error]( https://github.com/nicolaiarocci/cerberus/blob/master/cerberus/cerberus.py#L274) if a `purge_unknown=True` was provided (mind the schema-validation, allow and purge must not be True at the same time).

be aware that you have to fetch the result from `Validator.document` after validation.

funkyfuture:
sorry, my mistake, the validator-configuration doesn't relate to schema-validation. and on a second thought, i think it's okay to allow any combination:

- `allow_unknown` - throw an error or not
- `purge_unknown` - always silent

misja:
`del self.document[field]` operates on a copy of the original document so will not work :(

funkyfuture:
yep, that's why you have to fetch the `document`-property from the used `Validator`-instance afterwards. `Validator.validate` itself only returns a boolean.

the copying of the document has been introduced just recently to provide a non-destructable mechanism for 'value-`coerce`ing'.

misja:
Exactly, the only obvious way I saw how to get to the document was to [wrap] (https://gist.github.com/misja/d9b88ea849b9ab86af95) `Validator.validate`, run the validation and then get the validated/coerced etc. document - afterwards as you say. Then loop through the `unknown field` errors together with the document and remove fields where necessary, including updating the errors dict.

funkyfuture:
i don't get it. can you post some snippet of the code that uses the validator?

misja:
Sure, you can see what I've done here [https://gist.github.com/misja/d9b88ea849b9ab86af95](https://gist.github.com/misja/d9b88ea849b9ab86af95)

funkyfuture:
yes, i saw that. i mean the code that actually calls a document-validation. what would hinder you to access the document-property in that context?

misja:
[...]

PR #101, opened 2015-05-25
Adds wrapper-method `validated`

funkyfuture:
takes a document as argument and returns a validated document or 'None' if validation failed as a helper for cases where the processed document is used

documentation updated

may partly solve #95


nicolaiarocci:
Cool.

nicolaiarocci:
Would be nice if we had a test covering this feature though.

funkyfuture:
done and rebased.

nicolaiarocci:
thanks!

Issue #105, opened 2015-06-08
% string formatting doesn't allow missing parameters


brettatoms:
Consider using str.format() methods so that if someone wants to override the error.* messages you don't get an error if the %s is missing new error message.

```
>>> 'some message' % 'x'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: not all arguments converted during string formatting
>>> 'some message'.format('x')
'some message'
>>>```

PR #106, opened 2015-06-08
use str.format

brettatoms:


brettatoms:
After creating #105 I remembered I had already done this a while back.

Issue #108, opened 2015-06-10
Request: Allow list of schema to be validated against


CD3:
It would be useful to allow the `schema` key and  to be a list, which would be interpreted as a set of possible schemas that should be applied and the entry should validate if any of the schemas validate. It is already possible to give a list of types, but if I want to make sure that a value is a dict that has one of two possible formats, it is not possible without creating _validate_type_* functions.

[validictory](http://validictory.readthedocs.org/en/latest/usage.html#specifying-custom-types) allows this because it handles custom types in a way to the proposal in Issue [#96](https://github.com/nicolaiarocci/cerberus/issues/96)

funkyfuture:
i think an example would be useful.

would that be an equivalent to the items-constraint in a list's context?

CD3:
I'd like to be able to do something like this (or equivalent)

```
      schema   = { 'parts' : { 'type' : 'list'
                             , 'schema' :
                               { 'type' : [ 'dict', 'string' ]
                               , 'schema' :
                                 [
                                   { 'model number' : {'type' : 'string', 'required' : True}, 'count' : {'type' : 'integer'}  }
                                  ,{ 'serial number' : {'type' : 'string', 'required' : True}, 'count' : {'type' : 'integer'}   }
                                 ]
                               }
                            }
                 }
      v = Validator(schema)
      document = { 'parts' : [ 
                              { 'model number'  : 'MX-009', 'count' : 100 }
                             ,{ 'serial number' : '898-001', 'count' : 20 }
                             ,  'misc'
                             ]
                 }

      res = v.validate( document )

```

so 'parts' is allowed to be a string or a dict. If it is a dict, it either has to have a 'model number' entry or 'serial number' entry.

I don't think it items-constraint works here. My understanding is that the items-constraint makes sure that the items in a list (in the document) each validate to some schema.

funkyfuture:
> would that be an equivalent to the items-constraint in a list's context?

i reviewed the documentation and found that also these constraints beside *schema* and *type* could be used with a set of rules that are evaluated as an or-condition:

- dependencies
- items
- propertyschema
- regex
- schema (dict)
- valueschema

therefore is suggest a more generic solution that allows or-evaluated constraints.

CD3:
I agree, this would be useful.

Wouldn't there need to be another constraint added, for example 'anyof', to handle this transparently? You could store a list of or'able constraints, but then user defined validation functions would not have have access to them. What about allowing something like this:

```
[...]

Issue #107, opened 2015-06-09
[BUG] Nested document validation is broken


vinilios:
Nested document validation seems to be broken in master branch. After a bit of debugging it seems that this is due to the fact that self.document attribute of subdocument validators gets assigned with a copy of the contents of the parent document (`context`) causing further validation steps to fail.

To reproduce

``` python
import cerberus

schema = {
    'info': {
        'type': 'dict',
        'schema': {
            'name': {'type': 'string', 'required': True}
        }
    }
}

validator = cerberus.Validator(schema)
res = validator.validate({'info': {'name': 'my name'}})
if not res:
    print validator._errors

```

funkyfuture:
hm, [this](https://github.com/funkyfuture/cerberus/commit/6f68def9fdb289abe93fa40373d6ee64f2c78870) doesn't fail.

what's actually the output of your code above?

vinilios:
The _errors include the following,
```
{'info': {'name': 'required field'}}
```

This seems to got introduced with f66d8bd4880ddb7dfd443f5aea1dccb1152052d5. This commit is not in your fork though and i guess that's why you don't get an error.

CD3:
I'm seeing the same problem.

I think the problem is in the __validate_required_fields function. There are two lines that get the set of missing fields.

```
required = list(field for field, definition in self.schema.items()
                      if definition.get('required') is True)
missing = set(required) - set(key for key in document.keys()
                                      if document.get(key) is not None or
                                      not self.ignore_none_values)
```

`self.schema` is the subdocument schema, but `document` is always coming in as the full dict being validated. So, the list of required keys is generated correctly for each level in a document, but then the keys in the top level of the document are checked instead of the keys at the correct level.

CD3:
If I revert `self.document` back to `document`, undoing commit `f66d8bd`, it fixes the error.

funkyfuture:
well, i was the one that authored the commit in question. so, yes i'm actually using this code. can you please cherry-pick [this commit](https://github.com/funkyfuture/cerberus/commit/6f68def9fdb289abe93fa40373d6ee64f2c78870) and run the tests? it should fail. if not, please investigate.

i'm, too tired to hack around now.
but as @CD3 points at some interesting piece of code, my *guess* is that it could be rather caused because `self.document` *is not* referenced, but `document`.

here's my git log for comparison:
```
6f68def Adds a test for #107
e680644 Merge branch 'use-str.format-for-error-messages'
[...]

Issue #84, opened 2015-04-26
Proposal: add exclusion-checks


funkyfuture:
it'd be handy to declare properties to exlude others:

```python
{'property_a': {'excludes': 'property_b'},
 'property_b': {'excludes': ['property_a', 'property_c']}
```

if a property is not present in the document, excluded by a present one, but marked required, the requirement should be dismissed. this way mutual exclusive requirements are possible:

```python
schema = \
{'property_a': {'excludes': 'property_b', 'required': True},
 'property_b': {'excludes': 'property_a', 'required': True}}

valid_documents = \
[{'property_a': 'foo'}, {'property_c': 'bar'}]

invalid_documents = \
[{'property_a': 'foo', 'property_b': 'bar'}, {'property_c': 'baz'}]
```

since i can make use of this and may be implementing the next days, i'd appreciate any thoughts on this.

nicolaiarocci:
Looks good to me!

funkyfuture:
eventually it turned out, that this wouldn't be of help for me for now.

so, maybe later, but certainly not within the next days. so anyone who feels the urge, feel free to pick it. shouldn't be to tricky either.

funkyfuture:
wouldn't it be a good idea to keep this issue open?

funkyfuture:
i have the *feeling* that my use-case may be covered with the *of-rules. however i still like this more obvious way.

funkyfuture:
solved by #137. :rocket:

PR #109, opened 2015-06-11
Added support for lists of schema (#108)

CD3:
Schemas can now be specified as a list, in which case each schema in the
list will be tried and if one schema validates, the schema will be
considered as validated.


cerberus/cerberus.py
```diff
@@ -208,6 +208,37 @@ def _validate(self, document, schema=None, update=False, context=None):
...
+        # if self.schema is a list, then we want to validate the document against each item
+        # in the list and return true if one schema in the list validates.
+        if isinstance( self.schema, Sequence ) and not isinstance(self.schema, _str_type):
+            # assume that we no schemas validate
+            valid = False
+            # keep an error stack to pass on in case no schemas validate
+            errorstack = []
+            # for each schema...
+            for s in self.schema:
+                validator = copy.copy(self)
```
funkyfuture: better use `Validator.__get_child_validator(schema=s)`

funkyfuture:
one question that comes to my mind: how many other constraints could be used with a one-of-many-scenario. if there are some, it should be implemented in a generic way.

cerberus/cerberus.py
```diff
@@ -208,6 +208,37 @@ def _validate(self, document, schema=None, update=False, context=None):
...
+                validator.schema = s
+                validator.validate(document
+                                 , context=context
+                                 , update=update)
+                if len(validator.errors):
```
funkyfuture: `if validator.errors:` is more pythonic.

cerberus/cerberus.py
```diff
@@ -208,6 +208,37 @@ def _validate(self, document, schema=None, update=False, context=None):
...
+                    errorstack.append( validator.errors )
+                else:
+                    # no errors occured, set valid and get out of here
+                    valid = True
+                    break
+
+            if not valid:
+                # no schema in the list validated.
+                # add all the errors we have collected to the errors list
+                for i in xrange(len(errorstack)):
```
funkyfuture: you can simply `self._errors.update(validator.errors)`

cerberus/cerberus.py
```diff
@@ -208,6 +208,37 @@ def _validate(self, document, schema=None, update=False, context=None):
...
+                    self._error( "candidate %d"%i, errorstack[i] )
+
+            return valid
```
funkyfuture: this stops the validation regardless of further constraints.

funkyfuture:
wouldn't [exclusions-checks](https://github.com/nicolaiarocci/cerberus/issues/84) be more efficient (in terms of less code) in the scenario that is given in the tests?

cerberus/cerberus.py
CD3: Got it, I copied the loop from `_validate_schema` function. If `value` is a `Mapping`, the `copy.copy(self)` is used. Perhaps that should be changed to?

cerberus/cerberus.py
CD3: Got it, same as above.

cerberus/cerberus.py
CD3: Actually, we can't. If we add errors to the `_errors` list, then we will raise an exception later on even if one of the schemas validated. We only want to add the errors if validation does not occur.

cerberus/cerberus.py
CD3: Yes, if one of the schemas validates, we consider the document to be valid and there is no need to check the rest. This only runs if the schema is a `list`, so it will return control back to the parent validator which will continue checking constraints.

CD3:
Having one-of-many support for all constraints would be useful, but I will have to think about a general solution more.

The place to do this would be in `_validate_schema`. There is a loop that runs through all constraints that starts out like this.

```
        for field, constraints in schema.items():
            if not isinstance(constraints, Mapping):
                raise SchemaError(errors.ERROR_DEFINITION_FORMAT.format(field))
            for constraint, value in constraints.items():
                if constraint == 'type':
                    values = value if isinstance(value, list) else [value]
                    for value in values:
                        if not hasattr(self, '_validate_type_' + value):
# and so on...
```

So the 'type' constraint loops through a list of types of `values` is a list. It might be possible to just move this loop outside of the `if constraint == 'type'` block, but it may not work because I believe there are some constraints that already take a list. So these would have to be handled as special cases to make sure their lists where passed through.

Well, the more I think about it, I don't think we can do that. A user could define a constraint function that expects a list as its argument, so the validator should not break up lists before passing them to validator.

CD3:
exclustions-check might work in this specific case, but test I wrote is contrived. I want to be able to validate against a set of arbitrarily complex schemas and have validation pass if one passes.

CD3:
Just pushed updates

nicolaiarocci:
Thanks this looks like a nice addition. A few things needs to be addressed though:

- Under Python 3.4 tests are failing because `xrange` does not exist in Py3. Actually the contrary is true: in Python 3 `range` *is* `xrange`. You might want to use `range` for a quick fix: very efficient on Py3, less so on Py2 which is acceptable. The alternative is a test of the python version (which we already have for string types).
- There's a failure on the first validation attempt (https://travis-ci.org/nicolaiarocci/cerberus/jobs/66442227) 
- flake/pep issues (https://travis-ci.org/nicolaiarocci/cerberus/jobs/66442232)

cerberus/cerberus.py
```diff
@@ -208,6 +208,36 @@ def _validate(self, document, schema=None, update=False, context=None):
...
+                    valid = True
+                    break
+
+            if not valid:
+                # no schema in the list validated.
+                # add all the errors we have collected to the errors list
+                for i in xrange(len(errorstack)):
+                    self._error( "schema %d"%i, errorstack[i] )
+
+            return valid
```
funkyfuture: this makes the validation blocking (thus not checking the other constraints), but a design-paradigm is that cerberus is a non-blocking validator.

CD3: Well, that's embarrassing, the only test that fails is the test I wrote to demonstrate it work! Let me look into it. This may be related to [107](https://github.com/nicolaiarocci/cerberus/issues/107). I am using a patch that is not in the tree because I see the same error that was reported but funkyfuture does not.

CD3: No, I don't think it is blocking. In this case, there is a list of candidate schema. If one schema in the list validates, the constraint is considered satisfied.

The return statement will only get called when the schema is a list. For each item in the list, the same non-blocking validation is applied.

At least this is what should happen. I'll write a unit test to make sure that is was is happening.

funkyfuture: yeah, makes sense when i consider all you changes. ;-)

CD3:
Flake should pass now.

The unit tests fail because of issue #107. I have revised the tests so that they don't use nested 'required' constraints, but still test this feature. They are passing on my machine now.

CD3:
I have implemented a more general solution to produce this functionality that is simpler. Will submit a pull request for that branch instead.

PR #111, opened 2015-06-13
Issue 107

funkyfuture:
this this still WIP, the PR is intended to run travis' tests and discuss changes.
eventually it shall solve #107.

state of progress:

i haven't yet figured out why the tests do not fail when `assertSuccess` is used. but i think this is an important issue.

i can confirm that referring `self.document` in cerberus.py#302 makes a difference.

the failure was then caused by the fact, that in `_validate_schema` the validator copied itself and the copy then changed its schema-property which would then be reflected in the first validator, that would then continue validation with the subschema, causing the failure as the `required` field wouldn't exist in the higher context.
therefore i changed the code, so a new instance of a validator was used, not a copy. (on that occasion i made sure a child-validator would inherit all `__init__`-parameters unless overwritten.

now the problem is that in cerberus.py#227 a given `context` is always given precedence. which will fail if a subschema is to be validated. but omitting the context (cerberus.py#542) will lead to other failues.

@nicolaiarocci i'd appreciate some advice at this point. especially the code in cerberus.py#227 seems rather strange to me. why would there be a `context`-parameter which always supersedes the `document`-parameter? the calling code could / should figure what is to be validated.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -130,16 +130,25 @@ class Validator(object):
     special_rules = "required", "nullable", "type", "dependencies", \
                     "readonly", "allow_unknown", "schema", "coerce"
 
-    def __init__(self, schema=None, transparent_schema_rules=False,
-                 ignore_none_values=False, allow_unknown=False, **kwargs):
-        self.schema = schema
-        self.transparent_schema_rules = transparent_schema_rules
-        self.ignore_none_values = ignore_none_values
-        self.allow_unknown = allow_unknown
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
`context` was introduced by @joshvillbrandt in #42 which, in turn, was mainly introduced because of nicolaiarocci/eve/issues/295. I don't have time to look into it right now; hopefully I will in a few days. I'm fine with refactoring of course, as long as we don't introduce regressions. 

Josh and/or @nckpark might want to chime in.

Issue #93, opened 2015-05-22
Allow to pass error-handlers to a `Validator`-instance


funkyfuture:
this is a follow-up to #89 and #90.

i propose to introduce error-handlers in order to allow to deal with errors more flexible.

to achieve that:

- `Validator.__init__` takes an optional `error_handler`-object
- `Validator._error` is extended, so it stores the following data about an error:
  - `trail` - a list that represents the path to the `field` in the document (eg: `['a_dict', 'a_list']`); a common prefix can be specified upon calling `Validator.validate`
  - `field`, `value` - as of now
  - `constraint` - the constraint that failed
  - `message` - a simple error message, like currently implemented
- `Validator.errors` calls the `format`(?)-method of `error_handler` that may return errors in a desired format and / or do whatever its purpose is

there will be one default-handler and two as reference-implementations:

- `BasicErrorHandler`
  - returns errors as now
  - but concentates `trail` and `field`

- `HumanReadableErrorHandler`
  - is targeted to end-users
  - concentates `trail` and `field`
  - [suggests valid keys](https://github.com/nicolaiarocci/cerberus/issues/90), in case of an unallowed value
  - the index of list-items is increased by one
  - list-items are prefixed with `item #`

- `YamlErrorHandler`
  - structures errors in a dictionary
  - joins it into a yaml-file

funkyfuture:
i'm willing to implement this, but will complete other TODOs before. any feedback is appreciated.

CD3:
I think this is a great idea. I am using YAML to display error messages in my application, it is just a lot simpler to see what the problem is.

nicolaiarocci:
Like it. Scheduling for 0.10 release.

funkyfuture:
i'm currently working on that whenever i have the time and hope to provide a peek on wip this week (next week i'll be abroad).
considering how much unreleased changes currently cumulated and the amount of changes the error-handling will introduce, i'd say it'd be a good idea to postpone this to a `0.11`-release.

funkyfuture:
closing this as the core interface is now available.

Prudhvee:
Is `HumanReadableErrorHandler` implemented? Couldn't find the documentation.

PR #112, opened 2015-06-14
Feature anyof

CD3:
Adds an 'anyof' rule that will allow a list of definitions to be given. Currently, all rules in a definition are checked, so a rule is a set of constraints that are AND'ed together. The 'anyof' key allows constraints to be OR'ed. This should work transparently will all existing rules and any user defined validators.

Simple example, taken from unit test
```
# prop1 must be either a number between 0 and 10
schema = {'prop1':  {'min': 0, 'max': 10}}
doc = {'prop1': 5}

self.assertSuccess(doc, schema)

# prop1 must be either a number between 0 and 10 or 100 and 110
schema = {'prop1':
                  {'anyof':
                   [{'min': 0, 'max': 10}, {'min': 100, 'max': 110}]}}
doc = {'prop1': 105}

self.assertSuccess(doc, schema)

# prop1 must be either a number between 0 and 10 or 100 and 110
schema = {'prop1':
                  {'anyof':
                   [{'min': 0, 'max': 10}, {'min': 100, 'max': 110}]}}
doc = {'prop1': 50}

self.assertValidationError(doc, schema)

```

This also satisfies #108. See the [unit test](https://github.com/CD3/cerberus/commit/53c6b72b90bf4827038252a29fc8be4f11a1a938) for the syntax.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -229,50 +229,7 @@ class Validator(object):
 
             definition = self.schema.get(field)
             if definition is not None:
-                if value is None:
-                    if definition.get("nullable", False) is True:
-                        continue
-                    else:
-                        self._error(field, errors.ERROR_NOT_NULLABLE)
-
-                if 'coerce' in definition:
-                    value = self._validate_coerce(definition['coerce'], field,
-                                                  value)
-                    self.document[field] = value
-
-                if 'readonly' in definition:
-                    self._validate_readonly(definition['readonly'], field,
-                                            value)
-                    if self.errors.get(field):
-                        continue
-
-                if 'type' in definition:
-                    self._validate_type(definition['type'], field, value)
-                    if self.errors.get(field):
-                        continue
-
-                if 'dependencies' in definition:
-                    self._validate_dependencies(
-                        document=self.document,
-                        dependencies=definition["dependencies"],
-                        field=field
-                    )
-                    if self.errors.get(field):
-                        continue
-
-                if 'schema' in definition:
-                    self._validate_schema(definition['schema'],
-                                          field,
-                                          value,
-                                          definition.get('allow_unknown'))
-
-                definition_rules = [rule for rule in definition.keys()
-                                    if rule not in self.special_rules]
-                for rule in definition_rules:
-                    validatorname = "_validate_" + rule.replace(" ", "_")
-                    validator = getattr(self, validatorname, None)
-                    if validator:
-                        validator(definition[rule], field, value)
+                self._validate_definition(definition, field, value)
             else:
                 if self.allow_unknown:
                     if isinstance(self.allow_unknown, Mapping):
@@ -294,6 +251,52 @@ class Validator(object):
             self._validate_required_fields(self.document)
 
         return len(self._errors) == 0
+
+    def _validate_definition(self, definition, field, value):
+        if value is None:
+            if definition.get("nullable", False) is True:
+                return
+            else:
+                self._error(field, errors.ERROR_NOT_NULLABLE)
+
+        if 'coerce' in definition:
+            value = self._validate_coerce(definition['coerce'], field,
+                                          value)
+            self.document[field] = value
+
+        if 'readonly' in definition:
+            self._validate_readonly(definition['readonly'], field,
+                                    value)
+            if self.errors.get(field):
+                return
+
+        if 'type' in definition:
+            self._validate_type(definition['type'], field, value)
+            if self.errors.get(field):
+                return
+
+        if 'dependencies' in definition:
+            self._validate_dependencies(
+                document=self.document,
+                dependencies=definition["dependencies"],
+                field=field
+            )
+            if self.errors.get(field):
+                return
+
+        if 'schema' in definition:
+            self._validate_schema(definition['schema'],
+                                  field,
+                                  value,
+                                  definition.get('allow_unknown'))
+
+        definition_rules = [rule for rule in definition.keys()
+                            if rule not in self.special_rules]
+        for rule in definition_rules:
+            validatorname = "_validate_" + rule.replace(" ", "_")
+            validator = getattr(self, validatorname, None)
+            if validator:
+                validator(definition[rule], field, value)
 
     def _error(self, field, _error):
         field_errors = self._errors.get(field, [])
@@ -371,6 +374,20 @@ class Validator(object):
                             errors.ERROR_SCHEMA_TYPE.format(field))
                 elif constraint in self.special_rules:
                     pass
+                elif constraint in ('anyof', 'allof'):
+                    if(isinstance(value, Sequence) and
+                       not isinstance(value, _str_type)):
+                        # make sure each definition in an
+                        # anyof/allof constraint validates
+                        for v in value:
+                            # get a copy of the schema with
+                            # anyof/allof replaced with their value
+                            s = copy.copy(constraints)
+                            del s[constraint]
+                            s.update(v)
+                            self.validate_schema({field: s})
+                    else:
+                        self.validate_schema({field: [value]})
                 elif constraint == 'items':
                     if isinstance(value, Mapping):
                         # list of dicts, deprecated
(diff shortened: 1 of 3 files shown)


cerberus/cerberus.py
```diff
@@ -643,6 +646,25 @@ def _validate_validator(self, validator, field, value):
...
+            anyof = [anyof]
+
+        valid = False
+        tmperrors = self._errors
+        errorstack = {}
+        for i in range(len(anyof)):
+            definition = anyof[i]
+            self._errors = {}
+            self._validate_definition(definition, field, value)
+            errorstack["definition %d" % i] = self._errors.get(field, {})
```
funkyfuture: shouldn't that be after an `else`-statement after the next `if`!?

cerberus/cerberus.py
```diff
@@ -643,6 +646,25 @@ def _validate_validator(self, validator, field, value):
         # call customized validator function
         validator(field, value, self._error)
 
+    def _validate_anyof(self, anyof, field, value):
+        if isinstance(anyof, Mapping):
```
funkyfuture: is that to allow this?:
```python
schema = {'property': {'anyof': {one_rule}}}
```

CD3: Yes

funkyfuture:
looks good, but the problem is that you can't have multiple `anyof` in one schema. which you would need if you wanted to `and` two `anyof`-sets. but i can't come up with a pragmatic use-case and it seems obvious how a `allof` would be implemented.

i would undo any changes concerning the usage of `type`-rule with a list of valid types. does anyone disagree?

you may consider to squash your commits into one.

CD3:
hmm, good point. I'm not sure logical OR and AND commute or not.

I'll squash the commits.

What do suggest. Should I implement an `allof` an create a new pull requrest or add it to this one?

I think that the `type`  revert should be handled separably, it will probably require existing unit tests to be modified. Although, the current syntax may be a useful shorthand since wanting a list of types is clearly a common thing.

funkyfuture:
> What do suggest. Should I implement an allof an create a new pull requrest or add it to this one?

i'm not sure if this is actually needed in the real world.

also, i'm still thinking about coercion in that context. not sure, if it's a good idea to allow this.

what i meant: as i'm the one who did the changes regarding `type`, i would also undo them. but maybe there are objections.

btw, it *could* (didn't test it, just played around) be done in three lines less code:
```python
errorstack = {}

for definition, i in anyof, len(anyof):
    v = self.__get_child_validator(schema={field: definition})
    if v.validate({field}):
        valid = True
        break
    else:
        errorstack["definition %d" % i] = v.errors
    
if errorstack:
   self._error(field, errorstack)
```

CD3:
well, if that works it would not require the `validate_definition` function, which would be nice. let me try it out.

funkyfuture:
don't forget inline- and end-user-documentation. :-D

CD3:
of course, I'm new.

funkyfuture:
`Validator.validate_schmema` should ensure that only `Sequence`s of `Mapping`s are used.

CD3:
Well, using a child validator complicates things. The child validator runs the schema through `validate_schema` which fails on my unit test.

funkyfuture:
what's failing exactly?

CD3:
It doesn't like that a `schema` rule exists without a `type` key specifying `dict` or `list`. So it wants to see

```
{ 'anyof' : [ {'type' : 'dict', 'schema' : {...}}, {'type' : 'dict', 'schema' : {...}}]
```

instead of
```
{ 'type' : 'dict', 'anyof' : [ {'schema' : {...}}, {'schema' : {...}}]}
```

I didn't look at it anymore after I figured out that was what was wrong. The `_validate_anyof` function would need access to the level above its argument, which would require a workaround.

I'd actually like to change the `validate_schema` function to allow this,
```
{ 'type' : 'dict', 'schema' : { 'anyof' : [ {...}, {...}]}}
```
right now it complains that entries in 'schema' cannot be lists.

nicolaiarocci:
Hey guys, just wanted to let you know that I'm following the discussion and I like what you are doing here. 

> right now it complains that entries in 'schema' cannot be lists.

You could amend `validate_schema` to allow for `anyof`, it would not break backward compatibility I think.

CD3:
Thanks @nicolaiarocci, I actually wondered if I should stop looking at this until I heard from you.

A couple of things.

I added a `allof` constraint that does the same thing as `anyof`. Who knows, it may be useful and it seems more symmetric to me.

I also updated `validate_schema` to handle `anyof` and `allof` constraints. Basically, if one of these constraints is run into, it check that the schema formed by replacing the `*of` constraint with each of its values is valid. For example, if you have

```
schema = {'doc' : {'type' : 'number', 'anyof' : [ {'max' : 0}, {'min' : 10} ] } }
```
 then `validate_schema` would check that both
```
schema = {'doc' : {'type' : 'number', 'max' : 0 } }
```
and
```
schema = {'doc' : {'type' : 'number', 'max' : 0 } }
```
are valid schemas.

The addition to `validate_schema` failed the Trivis CI, I used dict comprehension which is not supported in 2.6.

@funkyfuture, I don't think you should remove list support from type constraints yet, `anyof` does not quite duplicate its behavior. If you wanted to specify multiple types, you have to do this

```
schema = {'parm' : {'anyof' : [ {'type : 'string'}, {'type' : 'number'} ] }
```

Its not possible to do

```
schema = {'parm' : {'type' : { 'anyof' : ['string', 'number'] } } }
```

Initially I thought it was because of the `validate_schema` function, but it is actually more than that. The function that validates `type` will be passed `{ 'anyof' : ['string', 'number'] } }`, which it does not know how to handle. The same is true for all other constraints too.

So, I think that adding support for prefixing keys with `anyof_` and `allof_` would also be useful (I think `anyof` constraint is still useful as is)

I'll fix the 2.6 compatibility and add user documentation if all agree that this feature will be accepted.

cerberus/tests/tests.py
```diff
@@ -850,6 +850,159 @@ def test_validated(self):
...
+
+        # prop1 must be either a number between 0 and 10 or 100 and 110
+        schema = {'prop1':
+                  {'anyof':
+                   [{'min': 0, 'max': 10}, {'min': 100, 'max': 110}]}}
+        doc = {'prop1': 50}
+
+        self.assertValidationError(doc, schema)
+
+        # prop1 must be either be less than 0 or less than 10
```
nicolaiarocci: you probably mean "equal or greater than 0 or equal or greater than 10"

cerberus/tests/tests.py
```diff
@@ -850,6 +850,159 @@ def test_validated(self):
...
+        schema = {'prop1': {'allof': [
+                 {'type': 'float'}, {'min': 0}, {'max': 10}]}}
+        doc = {'prop1': -1}
+        self.assertFail(doc, schema)
+        doc = {'prop1': 5}
+        self.assertSuccess(doc, schema)
+        doc = {'prop1': 11}
+        self.assertFail(doc, schema)
+
+        # prop1 has to be a float and a string
```
nicolaiarocci: "float and an integer"

cerberus/tests/tests.py
```diff
@@ -850,6 +850,159 @@ def test_validated(self):
...
+        self.assertSuccess(doc, schema)
+        doc = {'prop1': -1.5}
+        self.assertSuccess(doc, schema)
+        doc = {'prop1': 5.5}
+        self.assertFail(doc, schema)
+
+        doc = {'prop1': '5.5'}
+        self.assertFail(doc, schema)
+
+        doc = {'prop1': 5}
```
nicolaiarocci: test missing?

nicolaiarocci:
Green light from me. Make sure docs are updated appropriately. Tests are likely to serve as valid documentation examples too.

funkyfuture:
as you did the `allof`, we could complete the set of logical operators with `oneoff` (*one*, not more, must match) and `noneoff`.

when i'm home in some hours, i'd like to investigate shortly on that issue with a child-validator.

cerberus/cerberus.py
```diff
@@ -371,6 +374,20 @@ def update_to_valueschema(schema, warning_printed=False):
                             errors.ERROR_SCHEMA_TYPE.format(field))
                 elif constraint in self.special_rules:
                     pass
+                elif constraint == 'anyof' or constraint == 'allof':
```
funkyfuture: `elif constraint in ('anyof', 'allof'):` is shorter and better extendable.

funkyfuture:
> I don't think you should remove list support from type constraints yet, anyof does not quite duplicate its behavior.

i know, but i'd prefer one way to do it, and that would be the generic one.

please add a note to the documentation that usage of these rules is not tested in conjunction with coercion and not recommended.

funkyfuture:
fyi, as i would propably refactor to a usage of another instance as sketched above when i'm going to implement #93 anyway i pulled your code and tried it.

i also have that failing schema-validation on one of your tests (`test_anyof_schema`) as well. strangely when i follow the execution with a debugger, the code that will later be displayed as raising a `SchemaError` is actually not executed, and the `validate_schema`-method terminate normally.

CD3:
I think it is good to go. I won't have time to add `noneof` or `oneof` for a couple of days, but I'll work on it.

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/cerberus/commit/24f80286a4da33adff691e71ba3044939089aa26

Thanks!

PR #113, opened 2015-06-16
Keep a reference to the current document, 'rename' and 'remove_unknown' options

misja:
This relates to #95 , where no reference is kept to the document being validated and hence not being able to change or remove fields. This commit maintains references and adds a `current` property to the validator, pointing to the current document and making it available to custom validators.


cerberus/cerberus.py
```diff
@@ -701,6 +722,13 @@ def _validate_allof(self, allof, field, value):
         if not valid:
             self._error(field, errorstack)
 
+    def _validate_rename(self, rename, field, value):
+        if isinstance(rename, _str_type):
```
funkyfuture: this should be caught upon schema-validation.

funkyfuture:
there should be a `remove_unknown`-rule, similar to `allow_unknown`.

cerberus/cerberus.py
```diff
@@ -298,6 +316,9 @@ def _validate_definition(self, definition, field, value):
             if validator:
                 validator(definition[rule], field, value)
 
+        if 'rename' in definition:
```
funkyfuture: shouldn't this happen before further validation like coercion?

misja: I've purposely added it last since a rename will change a document field which could break other (custom/user) validators looking for the original field name.

misja: `remove_unknown` is included, see https://github.com/misja/cerberus/blob/references/cerberus/cerberus.py#L134

cerberus/cerberus.py
misja: I think schema validation will check the field value type only, the rename field is just a configuration option which should be a string since since its value will replace the original field name in the document.

funkyfuture: i don't get your point.
what hinders you from checking if the provided value in the schema is a string in `validate_schema`?

cerberus/cerberus.py
funkyfuture: i mean the possibility to override this option via a rule, that can be placed in a schema-defintion. since `allow_unknown` (as `Validator`-property) is inherited by derived validators, there is also the constraint `allow_unknown` in order to override this. please see the documentation about it.

cerberus/cerberus.py
misja: You're right, I'll add the check there and will change the type to `collections.Hashable`.

cerberus/cerberus.py
funkyfuture: > I've purposely added it last since a rename will change a document field which could break other (custom/user) validators looking for the original field name.

intuitively i'd expect that first a normalization is applied and the result is then validated.

misja: Makes sense, I'll put the rename first. As for adding a `remove_unknown` rule to the schema, will do.

funkyfuture:
as @misja [points out](https://github.com/nicolaiarocci/cerberus/issues/95#issuecomment-112426470) this actually solves #107. which is a very cool incident, so i can leave town w/o that issue open. i'm going to investigate and review more tomorrow regarding this issue.

an extension to the schema like the following works fine to check the correct execution order and result works fine:

    'propertyschema': {'type': 'string', 'regex': '^bar$'}

imo, the form `a_dict[…]` is more suitable in the tests than `a_dict.get(…)`. functionally there's nothing wrong about it, it's that cultural thing.

oh, what happens in a case like this?:
```python
schema = {'foo': {'type': 'string'},
          'bar': {'rename': 'foo'}}
document = {'bar': 42}
```

misja:
Right, I just did a final commit, just a tweak really. I did leave the `rename` at the end since changing the field name also requires modifying the schema during the validation, and frankly don't have the time for it anymore ... Same goes with adding `remove_unknown` to the schema - I tried but I keep losing the reference to the current document there. It will probably have to do with the field and values getting split everywhere to pass `field` and `value` to rules where it might probably better to pass the current document (`self.current` now) and `field`. It's essentially the difference between the following, where the latter will fail:

```python
x = {'root': {'child': 'my value'}}
y = x['root']
y['child'] = 'other value'
assert x == {'root': {'child': 'other value'}}
```
 ```python
x = {'root': {'child': 'my value'}}
y = x['root']['child']
y = 'other value'
assert x == {'root': {'child': 'other value'}}
```

**Comment by nicolaiarocci** [comment]
Cool stuff. 

One thing I'd do however, is make `current` a property, like `errors`, with proper self-documenting docstring. This way when docs are generated, devs take notice of this new feature.

**Comment by nicolaiarocci** [comment]
Also, for next time maybe, make sure you split every new feature or fix into individual pull requests. Makes it easier on the maintainer and more importantly on project history and therefore, for others to review.

**Comment by CD3** [comment]
I have been thinking about the 'reference to current document' feature recently. My though was to keep a current "path", probably just a member named `_path` that contains the keys required to get from the top of the document down to the current document.

What do you think about doing it this way instead. For my own applications it would be useful to know where in the document validation is occurring, but I think it would have the added benefit of not having to keep the main and current documents in sync.

**Comment by funkyfuture** [comment]
i was also thinking to refactor some stuff. especially a segregation of normalization of a document as a whole *before* validation.

what you propose as `path` may well be what i intend as `trail` with regards to #93.

but i can look into things at earliest as July.

**Comment by CD3** [comment]
I think you are right. Perhaps the Validator class keeps a `trail` that is just deep copied into `_errors`?

**Comment by funkyfuture** [comment]
that's the plan.

**Comment by CD3** [comment]
nice. then I think it would be better to have a `__get_current_document` function that gives you a reference to the current document using the main document and `trail`.

**Comment by nicolaiarocci** [comment]
I want/need to wrap up a 0.9 release soon, ideally as soon as #107 is solved. 

I can merge this so it gets included with next release,  which I suspect would be ideal for @misja, or keep it on hold until further development is accomplished. Given @funkyfuture and @CD3 planned developments around this, it might be advisable to hold for now. Thoughts?

**Comment by misja** [comment]
@funkyfuture @CD3  as I see it, adding a layer to keep a trail is routing around a more persistent problem. One can perfectly iterate through a dict and keep pointers to elements within the document, it's just the current splitting up and passing of field and value to rules which messes things up.

**Comment by misja** [comment]
@nicolaiarocci #107 could be solved now by cherry picking my commits related to setting `current`. As for future developments and reference handling, that's up to @funkyfuture and @CD3 :)

**Comment by nicolaiarocci** [comment]
@misja I know you're going to hate me for asking but, would you please resubmit as separate PRs? `remove_unknown`;  `rename`, #107 fix and/or `current` property? I'm a lazy bastard.

**Comment by CD3** [comment]
Having a `trail` would allow validators to consider things like the current field's parents or values of fields at the level above the current field, which would be useful to me, but this is probably a rare need. Perhaps having copy of `current` would be more useful to more people. If it fixes #107 I like it.

**Comment by funkyfuture** [comment]
unfortunately stress grew more than anticipated the last days and i won't be in town for two weeks.

i'm not very fond of a release atm. here are some reasons:

as i understand such thing would not fail respectively the behaviour is not determinate depending on order in which fields are validated, which is not inuitive imo:
```python
schema = {'foo': {'type': 'string'},
          'bar': {'rename': 'foo'}}
document = {'bar': 42}
```

there's no way to change the behaviour concerning `remove_unknown` in subschemas.

in my impression the code is meanwhile wildly grown and some consolidating refactoring should be done before further features are added. several aspects have been spotted and noted in the recent discussions. few i haven't publicly articulated yet. unfortuantely i don't have the time now to comprehend all that.

[this](https://github.com/nicolaiarocci/cerberus/pull/114#issuecomment-112572153) is still unclear to me. as the list as argument for types hasn't been released yet, i'd rather remove it as with `anyof` there's now a more generic solution.

also, @CD3 [points out](https://github.com/nicolaiarocci/cerberus/pull/114#issuecomment-112576911), that `*of`-rules don't work as expected in any constellation. or is this solved? if not, it seems that some refactoring is a key part in tackling that.

if a release is necessary soon, i'd pledge to make it a `0.9-rc0`.

CD3:
This issue has been fixed. The unit tests added in [this commit](https://github.com/nicolaiarocci/cerberus/commit/2b951f1ed9e7a6033a4f4a2f8030461d89e176e5) demonstrates that both organizations will work.

funkyfuture:
thanks for the feedback.

misja:
@funkyfuture To be honest I'm fine if this pull request is put on hold, I'm at least glad part of it has helped  squash a related issue (serendipity ftw!). Whilst working on it doubt has crept in though if this library is suitable enough for my use, so am leaving it for now.

funkyfuture:
that leaves the question about the removal of checking a list of types from my list. that's not too hard, once it's decided. we can also postpone an 'agglutinating' syntax for `*of`-rules onto a later release.

CD3:
I agree (but I realize I've only been contributing for about a week). You should remove the list of types feature before the release so that you/we can implement a generic solution without having to worry about breaking backward compatibility. If you don't remove it now, you won't be able to remove it for a while.

PR #114, opened 2015-06-16
Adds noneof and oneof rule.

CD3:
Added 'noneof' rule which provides a list of definitions that should all
not validate.

Added 'oneof' rule which provides a list of definitions in which only
one should validate.

Added unit tests for oneof and noneof rules.

Updated documentation to include oneof and noneof rules.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -374,7 +374,7 @@ class Validator(object):
                             errors.ERROR_SCHEMA_TYPE.format(field))
                 elif constraint in self.special_rules:
                     pass
-                elif constraint in ('anyof', 'allof'):
+                elif constraint in ('anyof', 'allof', 'noneof', 'oneof'):
                     if(isinstance(value, Sequence) and
                        not isinstance(value, _str_type)):
                         # make sure each definition in an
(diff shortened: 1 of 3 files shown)


funkyfuture:
as interested in linguistics, it's cool to see the process of grammaticalization in action. :-D

i think with this logic-implemenation, i could easily try a replacement to the current possibility to validate against a list of types. `anyof_type = [type1, type2, typen]` would be translated to `anyof = [{'type': type1}, {'type': type2}, {'type': typen}]`. as a convenience in `validate_schema`.

not sure what's the linguistic category for this. agglutination may,

you can try first if you have the capacaties.

CD3:
Go ahead and take a crack on it. I am going to be offline for a while soon.

With respect to the logic implementation, I think it is necessary for `_validate_logical` to become a special rule.

The current implementation does not allow this
```
schema = {'prop': {'anyof': [ {'type': 'dict', 'schema': {...}},
                              {'type': 'dict', 'schema': {...}} ] } }
 ```
as I had originally thought. Instead, this is required
```
schema = {'prop': {'type': 'dict', 'anyof': [ {'schema': {...}},
                                              {'schema': {...}} ] } }
 ```
I believe the problem is that `schema` is a special key, and when `_validate` runs into it, it checks for the `type` in the value of `prop`.

So, if the result of an `*of` key was to just generate instances of schema where the `*of` key was replaced with the contents of each possibility, so that this
```
schema = {'prop': {'anyof': [ {'type': 'dict', 'schema': {'type' : 'integer'}},
                              {'type': 'dict', 'schema': {'type' : 'string'}} ] } }
 ```
turns into this
```
schema = {'prop': {'type': 'dict', 'schema': {'type' : 'integer'}}}
```
and this
```
schema = {'prop': {'type': 'dict', 'schema': {'type' : 'string'}}}
 ```
and the same thing happened for
```
schema = {'prop': {'type': 'dict', 'anyof': [ {'schema': {'type': 'integer'}},
                                              {'schema': {'type': 'string'}} ] } }
```
then you could use what ever syntax was convenient. I am thinking that it would be similar to what I did in the `validate_schema` function, where the `*of` entry is removed from the schema and replaced with each of its items one at a time with a dict update. This would allow a "common" rule for all candidates to be given, but then each candidate could overload it if they wanted.

But from what I can tell, this would require the `*of` rules to be aware of their surroundings, not just their children.

CD3:
OK, I have re-implemented the `*of` validation functions, it did not require them to be made "special".

@funkyfuture, this version uses your suggestion and adds support for the inline type syntax we have discussed.

nicolaiarocci:
Thanks

PR #117, opened 2015-06-18
Ensures to inherit properly to child-validators

funkyfuture:
- also adds a test for #107


nicolaiarocci:
Merged in https://github.com/nicolaiarocci/cerberus/commit/1a8540229b7f1ee600caba6fbf5a2fac7d709519

Thanks!

PR #119, opened 2015-06-19
Updates .travis.yml for faster builds

funkyfuture:
- use Docker under the hood
- cache pip's downloads between builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
+sudo: false
 language: python
+cache: pip
 python: 2.7
 env:
   - TOX_ENV=py26


nicolaiarocci:
In your experience, will this provide any real performance gain as we're using tox too? What I mean is, will it cache tox environments too?

funkyfuture:
what i expeienced with the build to this pr is what the travis team says: it's faster scheduled. the builds itself are not faster.

the caching affects only the pip-cache, so tox is not downloaded everytime.

nicolaiarocci:
Yeah that's what I thought.

PR #122, opened 2015-07-01
raises TypeError in case 'value' is a type object

alanderex:
small improvement:
if 'value' is a type object, the + concatenation raises a TypeError
'_validate_type_'.format(value) raises no error but still the correct
cerberus.cerberus.SchemaError: unrecognized data-type 'xyz'
is raised.


nicolaiarocci:
this breaks the current test suite. Can you fix it and rebase on top of current master?

PR #123, opened 2015-07-05
Refactoring and normalized-method

funkyfuture:
The main purposes of this PR is to prepare the code to implement #93 (and therefore rely completely on child-validators in order to follow object's trails). Furthermore it contains a lot of code-cleanup and -consolidation. some methods have been renamed for the sake of clarification.

~~**this isn't finished yet** as i will try to reduce code even more regarding one issue.~~

in a later commit i would reorder the methods to an order as meaningful as possible.

@nicolaiarocci if this is supposed to be included not before the 1.0-release, i would also remove any deprecated parts. as i would also like to apply that to the documentation, an older version of the docs may be made available publicly.

@misja i don't see the point of the `current`-method and would like to remove it. any objections?


funkyfuture:
eventually i remodeled the context's aspects that imo names the things in a clearer way.

however, i think it may even be useful to make what is now `Validator.context` available as `Validator.root_document` and hold the immediate parent-mapping as `Validator.context`.

funkyfuture:
@nicolaiarocci can you estimate when you will review this? i'd appreciate any short feedback for my planning.

funkyfuture:
> however, i think it may even be useful to make what is now Validator.context available as Validator.root_document and hold the immediate parent-mapping as Validator.context.

note to self: this may be easily tackled with a trail-property of Validator

cerberus/cerberus.py
```diff
@@ -157,20 +171,24 @@ def __init__(self, *args, **kwargs):
...
+        self.document = None
+        self.update = False
+        if self.schema:
+            self.validate_definition_schema(self.schema)
 
     def __call__(self, *args, **kwargs):
         return self.validate(*args, **kwargs)
 
+    # TODO REMOVE? isn't used in the code,
+    # seems uninteresting for the public api
```
nicolaiarocci: yup

nicolaiarocci:
Looks good to me - also a much needed refactor and cleanup work. We could merge this so `normalized` is available with 0.10. Cleanup of deprecated stuff can come up later, when 1.0 is released.

funkyfuture:
what about the reanimg of `self._current` to `self.document` and `self.document` to `self.context`. that could break custom validators and can't be aliased somehow.

i'd like to change that rather sooner than later as it clarifies a lot. or should i rather change this back and add comments as explanation and reminder?

nicolaiarocci:
That's one tough question. Since we're still in beta though, I'd go ahead with the change. Make sure changelog is properly updated. Maybe also add docstrings notes.

funkyfuture:
i named it `root_document`. that's unambigious.

but i didn't change it in `test_self_document_always_root`. and it doesn't fail. ?:-/

otherwise i'm fine with merging it.

funkyfuture:
i added a commit that reorders the methods and comes with a minor fix and docs extension.

can you think of a clearer name for `Validator.update`?

funkyfuture:
> but i didn't change it in test_self_document_always_root. and it doesn't fail. ?:-/
fixed it. my bad

PR #125, opened 2015-07-08
Adds the ability to extend of-rules with another rule

funkyfuture:
this adds convenience if you use one of the *of-rules with a set of single and identical rules.

PR #126, opened 2015-07-09
Adds testing with Docker

funkyfuture:
since #125 didn't trigger a build at Travis, i built a script that uses Docker to run all test locally against all interpreters without much hassle.


funkyfuture:
as the weblinks in the documentation are validated with the last commit, it is failing because the Google Groups-page can't be verified. they are using the `#` in a strange manner.
would you remove that link? something like whitlisting seems not to be possible with the doclinks-extension and i don't see some workaround in sight.
maybe it's even better to direct questions regarding usage towards Stackoverlow. anything bug- and feature-related is fine here.

nicolaiarocci:
Add `linkcheck_anchors` to conf.py to fix the google groups issue (see [docs](http://sphinx-doc.org/latest/config.html#confval-linkcheck_anchors)). I tested it on your PR and it works like a charm.

Also, can you please rebase on top of master?

funkyfuture:
thanks for the hint. i find that documentation tool's documentation still rather confusing. :-/

i now whitlelisted the google group's links, did you achieve successfully checking it?

also rebased it.

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/* \
+ && pyp install flake8 pytest tox PyYAML Sphinx \
+ && mkdir /home/tox \
+ && mv /root/.cache /home/tox/
+
+# this will be set to the user's id who's running a script that uses this
+# image in order to have file-ownerships on the host-system
+# at some point there should be native Docker-implementation for this
+ENV TOX_USER_ID=1000
```
nicolaiarocci: Is this something that needs editing at every first-time usage? If so, it should be added to the documentation for docker.

funkyfuture: no, it gets substitued in the `run-docker-tests` on each invokation to match file-ownerships with a possible `.tox`-folder.

nicolaiarocci: ah right, I see it now.

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/cerberus/commit/7ae46e57024d915f11502d8bde91ae51f3a1f510

funkyfuture:
ciao, is there a reason why you dismissed my commits and replaced it with your own?

nicolaiarocci:
Cant check right now I thought I pulled your rebase before the merge ? I had a rebased branch maybe I forgot to fetch the updates... Don't know will fix tomorrow if needed when I'm back.



—
Nicola Iarocci
http://nicolaiarocci.com

nicolaiarocci:
Your commits are there https://github.com/nicolaiarocci/cerberus/commits/master, it's just a merge --no-ff as usual.

funkyfuture:
okay, i wasn't sure if there was something wrong.

Issue #132, opened 2015-07-21
Multiple anonymous schema validation


calve:
Hi all,

I am looking for a schema that can validates multiples schema ; 

```
document1 = {
   "this_field" : {"bar_field" : "value"},
   "common_field1" : 42,
   "common_field2" : "hello",
}
document2 = {
   "that_field" : "blop",
   "common_field1" : 42,
   "common_field2" : "hello",
}
```

(note the difference between ``this_field`` and ``that_field``)

From what I understand from the documentation, it could be written using ``*-of``  :
```
schema = {
    "oneof_schema":[
        {"this_field" : {"type": "dict"}},
        {"that_field": {"type" : "string"}}
    ],
   "common_field1" : {"type" : "integer"},
   "common_field2" : {"type" : "string"}
}
```

When I try to validate, I get the following error 

```
>>> Validator().validate(document1, schema) 
 File "/home/g/.virtualenvs/p/lib/python2.7/site-packages/cerberus/cerberus.py", line 231, in validate
    return self._validate(document, schema, update=update, context=context)
  File "/home/g/.virtualenvs/p/lib/python2.7/site-packages/cerberus/cerberus.py", line 249, in _validate
    self.validate_schema(schema)
  File "/home/g/.virtualenvs/p/lib/python2.7/site-packages/cerberus/cerberus.py", line 397, in validate_schema
    raise SchemaError(errors.ERROR_DEFINITION_FORMAT.format(field))

cerberus.cerberus.SchemaError: unknown rule 'oneof_schema' for field 'oneof_schema'
```

I need to "name" the field used by `oneof` to pass this error

```
schema = {
    "named" : {
        "type" : "dict",
        "oneof": [
            {"schema": {"this_field" : {"type": "dict"}}},
            {"schema": {"that_field": {"type" : "string"}}}
        ]
    },
   "common_field1" : {"type" : "integer"},
   "common_field2" : {"type" : "string"}
}
```

but this won't validate my original ``document1`` and ``document2``, which now needs to contains a ``named`` key
```
document3 = {
   "named" : {"this_field" : {"foo":42}},
   "common_field1" : 34
}

document4 = {
   "named" : {"that_field" : "test"},
   "common_field1" : 34
}
```

Correctly outputs

```
>>> Validator().validate(document3, schema)
True
>>> Validator().validate(document4, schema)
True
```

But this is not as good as what I need since it changes the structures of my documents (``document1 != document3`` and ``document2 != document4``)

[...]

PR #135, opened 2015-08-08
Definition schema class

funkyfuture:
this includes a commit that improves DefinitionSchema and one that fixes an issue that came up in the doctests for the first time.

PR #137, opened 2015-08-19
Implements `excludes`

calve:
As suggested in #132.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -416,6 +416,7 @@ class Validator(object):
         self.update = update
         self.__init_processing(document, schema)
         self.__prepare_document(document, normalize)
+        self._required_excluded = set()
 
         for field in self.document:
             if self.ignore_none_values and self.document[field] is None:
@@ -707,12 +708,22 @@ class Validator(object):
         """
         required = set(field for field, definition in self.schema.items()
                        if definition.get('required') is True)
+        required -= self._required_excluded
         missing = required - set(field for field in document
                                  if document.get(field) is not None or
                                  not self.ignore_none_values)
 
         for field in missing:
             self._error(field, errors.ERROR_REQUIRED_FIELD)
+
+        # At least on field from self._required_excluded should be
+        # present in document
+        if len(self._required_excluded) > 0:
+            fields = set(field for field in document
+                         if document.get(field) is not None)
+            if self._required_excluded.isdisjoint(fields):
+                for field in self._required_excluded - fields:
+                    self._error(field, errors.ERROR_REQUIRED_FIELD)
 
     def _validate_schema(self, definition, field, value):
         schema = definition.get('schema')
@@ -821,6 +832,16 @@ class Validator(object):
     def _validate_excludes(self, excludes, field, value):
         if isinstance(excludes, Hashable):
             excludes = [excludes]
+
+        # Save required field to be checked latter
+        if 'required' in self.schema[field] and self.schema[field]['required']:
+            self._required_excluded.add(field)
+        for exclude in excludes:
+            if (exclude in self.schema
+               and 'required' in self.schema[exclude]
+               and self.schema[exclude]['required']):
+
+                self._required_excluded.add(exclude)
 
         if [True for key in excludes if key in self.document]:
             # Wrap each field in `excludes` list between quotes
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -818,6 +818,15 @@ class Validator(object):
                 if len(validator.errors):
                     self._error(field, validator.errors)
 
+    def _validate_excludes(self, exclude, field, value):
+        if isinstance(exclude, list):
+            for key in exclude:
+                if key in self.document:
+                    self._error(field, errors.ERROR_EXCLUDES_FIELD.format(exclude, field))
+        else:
+            if exclude in self.document:
+                self._error(field, errors.ERROR_EXCLUDES_FIELD.format(exclude, field))
+
 
 class DefinitionSchema(MutableMapping):
     """ A dict-subclass for caching of validated schemas.
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -1031,7 +1031,7 @@ class DefinitionSchema(MutableMapping):
         for key in excludes:
             if not isinstance(key, _str_type):
                 raise SchemaError(
-                    errors.ERROR_EXCLUDES_HASHABLE.format(key))
+                    errors.SCHEMA_ERROR_EXCLUDES_HASHABLE.format(key))
 
 
 def expand_definition_schema(schema):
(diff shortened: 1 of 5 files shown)


nicolaiarocci:
Excellent.

cerberus/cerberus.py
```diff
@@ -758,6 +758,17 @@ def _validate_valueschema(self, schema, field, value):
                 if len(validator.errors):
                     self._error(field, validator.errors)
 
+    def _validate_excludes(self, exclude, field, value):
+        targets = []
+        if isinstance(exclude, list):
```
funkyfuture: please test for `isinstance(exclude, Sequence) and not isinstance(exclude, _str_types)`

cerberus/cerberus.py
```diff
@@ -758,6 +758,17 @@ def _validate_valueschema(self, schema, field, value):
                 if len(validator.errors):
                     self._error(field, validator.errors)
 
+    def _validate_excludes(self, exclude, field, value):
```
funkyfuture: should be named `excludes` as a list is expected.

cerberus/cerberus.py
```diff
@@ -758,6 +758,17 @@ def _validate_valueschema(self, schema, field, value):
...
+        targets = []
```
funkyfuture: it's enough to wrap a single string into a list with the name `excludes`, no need for an extra object `target` imo.

funkyfuture:
please also add a check to `SchemaError.validate` for the `excludes`-rule. please look at #135 for the scheme i used to refactor schema-errors.

otherwise, lgtm. and i'm even baffled that `required` is dismissed correctly without extra effort. 8-)

docs/usage.rst
```diff
@@ -616,6 +616,62 @@ Thus you can use this to validate a document against several schemas without imp
...
+
+.. doctest::
+
+    >>> v = Validator()
+    >>> schema = {'this_field': {'type': 'dict',
+    ...                          'excludes': 'that_field'},
+    ...           'that_field': {'type': 'dict',
+    ...                          'excludes': 'this_field'}}
+    >>> v.validate({'this_field': {}, 'that_field': {}}, schema)
+    ... False
```
funkyfuture: that has to be

```
False
```

also true for all following output of returned objects. with leading dots it's interpreted as statement, not as expected result.

calve:
Thank you very much for the review. I considered your observations and pushed the appropriate changes.

However, I am not quite sure what you mean by 

> add a check to ``SchemaError.validate``

I added a test for the error string. Please elaborate if this was not what you meant.

funkyfuture:
ah, sorry. i meant [`DefinitionSchema.validate`](https://github.com/nicolaiarocci/cerberus/pull/137/files#diff-aebc53cd4926f3a579adb7ba188de369R809); ensure that the defintion for an `excludes`-constraint is either a Hashable or a (non-string-)sequence of `Hashable`s to avoid runtime-errors on validation.

funkyfuture:
though i don't see a use-case for it, you might add handling of the rule for unknown fields for the sake of completeness. that would let *any* arbritary field to exclude one or more defined fields.

cerberus/cerberus.py
```diff
@@ -758,6 +758,16 @@ def _validate_valueschema(self, schema, field, value):
                 if len(validator.errors):
                     self._error(field, validator.errors)
 
+    def _validate_excludes(self, excludes, field, value):
+        if isinstance(excludes, _str_type):
```
funkyfuture: should test for `Hashable` as any hashable object can be a mapping's key.

nicolaiarocci:
Could you please rebase this one of top of current `master`? Thanks!

calve:
Rebased.

Running ``tox`` test suite, those two errors are raised;

```
Document: usage
---------------
**********************************************************************
File "usage.rst", line 654, in default
Failed example:
    v.validate({'this_field': {}}, schema)
Expected:
    True
Got:
    False
**********************************************************************
File "usage.rst", line 656, in default
Failed example:
    v.validate({'that_field': {}}, schema)
Expected:
    True
Got:
    False
**********************************************************************
```

which are not consistent with the results expected by the test suite. Any idea how to fix those ?

funkyfuture:
in such cases  i adapt the doctests as unittests to investigate.

nicolaiarocci:
hmm I just ran tox and didn't get any issues (travis agrees too).

nicolaiarocci:
For line 654:

    >>> v.errors
    {'that_field': 'required field'}

nicolaiarocci:
The similar test succeeds because `assertSuccess` passes `update=True` to the `validate` method. If `update` is `False`, then it will fail (as it happens to be in the doctest).

calve:
It looks like when both fields are required, the validatiion always fails. That makes sence, but the following test case should fails.

    def test_required_excludes(self):
        schema = {'this_field': {'type': 'dict',
                                 'excludes': 'that_field',
                                 'required': True},
                  'that_field': {'type': 'dict',
                                 'excludes': 'this_field',
                                 'required': True}}
        document1 = {'this_field': {}}
        document2 = {'that_field': {}}
        document3 = {'that_field': {}, 'this_field': {}}
        self.assertSuccess(document1, schema)
        self.assertSuccess(document2, schema)
        self.assertFail({}, schema)
        self.assertFail(document3, schema)

in particular the following lines : 
    
        self.assertSuccess(document1, schema)
        self.assertSuccess(document2, schema)

Actually, the test suite still succeed if I changed those ``assertSuccess`` for ``assertFail``

nicolaiarocci:
Yup see my comment on `assertSuccess` above

calve:
This makes sense. Do you have a suggestion about it ?

I am about to patch the ``assertSuccess`` method to accept an ``update`` parameter.

nicolaiarocci:
Patching `assertSuccess` is probably a good thing. You will also want to edit/change the usage.rst example.

calve:
Please have a look at the latest commit : running ``assertSuccess`` with ``update=False`` raised errors concerning mutually excluded and required fields (in the ``test_required_excludes`` function). 

``required`` fields were not properly dismissed, which should be fixed now.

calve:
Thanks for the review once again. I applied your suggestions.

cerberus/cerberus.py
```diff
@@ -818,6 +830,28 @@ def _validate_valueschema(self, schema, field, value):
                 if len(validator.errors):
                     self._error(field, validator.errors)
 
+    def _validate_excludes(self, excludes, field, value):
```
funkyfuture: one thing i missed, within their functional grouped sections methods are sorted alphabetically. thus, this should be moved to line 578.

calve: Updated

nicolaiarocci:
Thanks!

nicolaiarocci:
PS: feel free to PR to replace your nickname if so you wish (could not get a hold of your name from the logs). Or I can edit it for you :)

funkyfuture:
great, thank you!

calve:
Thanks for all the reviews and the approval.

PR #140, opened 2015-09-06
Adds missing deprecation-warnings.

funkyfuture:
Also tries to obtain a logger, includes fallback to use `print`.


nicolaiarocci:
Merge with https://github.com/nicolaiarocci/cerberus/commit/e61fd8972c41366db513951184b830f55f133b7c

PR #143, opened 2015-09-08
some cleanup concerning child-validator's calls and tests

funkyfuture:
i started this to reduce code by directly calling child-validators which fits better into the 80 chars per line.
i also found a lot of potential to streamline tests.
making `Validator.__call__` an alias instead of a wrapper should slightly improve performance and is less error-prone.

please review carefully as i touched a lot of tests.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -10,6 +10,7 @@
 
 from collections import Callable, Hashable, Iterable, Mapping, MutableMapping, \
     Sequence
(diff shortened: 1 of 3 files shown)


nicolaiarocci:
Would you please rebase on top of current head? Thanks!

funkyfuture:
done.

nicolaiarocci:
Awesome as always. Thank you.

PR #152, opened 2015-09-26
bootstrap as theme

funkyfuture:
this is a preview of a documentation rendered with bootstrap as theme.

imo, the code parts are better readable, the navigation is more straightforward and it suits small and large screens thanks to responsiveness.

for now *there still is an issue* with the external links in the navigation bar; the urls get an `.html` appended.

i'm not sure why i had https://ghbtns.com/ successfully integrated at one point, but then it broke for yet unknown reasons.

i would also be much in favor of sorting the rules alphabetically for a better lookup.

### DIFF
--- a/docs/_templates/navbar.html
+++ b/docs/_templates/navbar.html
@@ -1,0 +1,49 @@
+<div id="navbar" class="{{ theme_navbar_class }} navbar-default {% if theme_navbar_fixed_top|tobool -%} navbar-fixed-top{%- endif -%}">
+  <div class="container">
+    <div class="navbar-header">
+      <!-- .btn-navbar is used as the toggle for collapsed navbar content -->
+      <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".nav-collapse">
+        <span class="icon-bar"></span>
+        <span class="icon-bar"></span>
+        <span class="icon-bar"></span>
+      </button>
+      <a class="navbar-brand" href="{{ pathto(master_doc) }}">
+        {%- block sidebarlogo %}
+          {%- if logo %}<img src="{{ pathto('_static/' + logo, 1) }}">{%- endif %}
+        {%- endblock %}
+        {% if theme_navbar_title -%}{{ theme_navbar_title|e }}{%- else -%}{{ project|e }}{%- endif -%}
+      </a>
+      <span class="navbar-text navbar-version pull-left"><b>{{ version|e }}</b></span>
+    </div>
+
+      <div class="collapse navbar-collapse nav-collapse">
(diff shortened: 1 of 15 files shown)


funkyfuture:
i'm ready with it.

i left out the github-buttons as this would need some extra styling which isn't simple to add atm. the theme would have to be improved before for a reasonable effort.

also, the references to the types in the Python documentation all point to Python 2's docs. i haven't yet figured out how to reference to Python 3's. i opened a support request regarding this.

nicolaiarocci:
hey thanks for this work.

Honestly, I am fond of the current theme and don't see a compelling reason to switch to the oh-so-commonly-used bootstrap theme. I do agree that navigation could be slightly improved: sorting the validation rules alphabetically is a good thing. I would keep the current theme, eventually updating navigation and/or sorting as needed.

funkyfuture:
i compiled some points that i considered as compelling reasons for me to address the html-layout.

on a 23" display (landscape):
- a code block with less than 120 chars is not full shown; on that same display my ide shows 2x120  and still has accessible navigations
- only about 40% of the width are used for actual content; that has an impact on height of content that makes browsing through a document an unsatisfying experience
- on the Usage-page which is most interesting as reference i am constantly hitting *pos1* and left-click in rotation

on a 10" tablet (portrait):
- on the Usage-page i find myself lost quiet soon; very much information in a more or less arbritrary order is the impression
- most code boxes would show their full content if space wasn't wasted for a nice white stripe on the left that makes only sense when you look at the top of a page; and it even doesn't fit a golden ratio

on a ~5" phone (portrait):
- starting with the main page, i have to scroll more than one screen down to get to the toc, which itself is four and a half screens long
- on each page one has to scroll to the bottom (try that on Usage) to have a toc at all, before that you're on youself and totally lost in finding a particular paragraph
- code blocks are in text boxes that are up to three widths broad; who can realize the actual code reading from that can consider her/himself a savant
- this way you can read the docs only as prose while ignoring the illustrations


since i was not keen to modernize alabaster, i chose to find a theme that is designed for various screen sizes and different orientations.
i'm not especially keen on bootstrap, but anticipated that this one is thouroughly developed due to its large userbase.
also. í found no alternative that also covers the issue of a sticky toc. and that is what makes a reference a modestly browsable and thus a far more usable in the meanings of a reference. even printed paper beats the status quo in this regard.

i'm sure that a well readable and navigatable documentation is a luxury that is worth a great library and should be as charming as the library itself.

nicolaiarocci:
Looks like there's an [attempt at making Alabaster more responsive](https://github.com/bitprophet/alabaster/issues/44). Hopefully the corresponding PR will be merged soon.

PR #162, opened 2015-10-14
Error system overhaul

funkyfuture:
this is my approach to tackle #93.

though it is still not fully ~~backwards-compatible (submitting custom errors is still missing, but seems rather trivial) and not~~ documented, this thing is at a point where i'd appreciate any review and feedback.

please do not merge yet.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,7 +10,7 @@ Not released yet.
 
 - New: 'excludes' rule (calve). Addresses #132.
 - New: Add testing with Docker (Frank Sachsenheim).
-- New: '*of'-rules can be extended by concentating another rule (Frank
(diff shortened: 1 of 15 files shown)


funkyfuture:
custom validators are now also backward-compatible.

funkyfuture:
a'ight, i added quiet some docs about the mess.

nicolaiarocci:
Stellar. Thanks!

funkyfuture:
oh, i think that could have waited. well, let's see what the world says.

i'll see that i elaborate one or two design questions that trouble me before moving forward with this in #93.

funkyfuture:
and thanks for the corrections.

PR #166, opened 2015-11-02
Adds document- and schema-error trees to Validator

funkyfuture:
This allows error-querying nodes through a container-interface.
And some smaller cleanup.

docs are still missing

### DIFF
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -2,7 +2,7 @@
 
 from collections import namedtuple
 from copy import copy
-from .utils import quote_string
+from .utils import compare_paths_lt, quote_string
 
 """
 Error definition constants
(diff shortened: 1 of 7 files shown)


funkyfuture:
~~this is fully functional now.~~
the most challenging part was actually to find a reasonable extent for the tests and user-documentation.

but, this also revealed two flaws (#169 and JSON doesn't know non-string keys in mappings). i will provide fixes for these and then rebase. shall i add them to this PR?

oh, and the tests are also failing on travis because list's order are not predictable. i will fix this too and tend to do this by changing error lists to be sorted (objections? thoughts on sorting criteria?).

anyway, a review of the rest is appreciated.

cerberus/errors.py
```diff
@@ -139,6 +148,113 @@ def is_logic_error(self):
...
+        self.descendants = dict()
+
+    def __add__(self, error):
+        self.add(error)
+        return self
+
+    def __delitem__(self, key):
+        del self.descendants[key]
+
+    def __iter__(self):
```
funkyfuture: would an iterator over `self.errors` be a more intuitive / usable convenience?

funkyfuture: /me thinks so
so i changed it.

funkyfuture:
from my perspective it's mergable.

Issue #131, opened 2015-07-20
Proposal: Add `default`-normalization rule


funkyfuture:
this is a follow-up of [this](https://github.com/nicolaiarocci/cerberus/issues/102#issuecomment-122241232).

A `default`-rule would substitute any `None`-values, empty Mapping and Sequence-instances and numbers equal to zero for a given field with the provided constraint's value while a document is normalized. if the field isn't present at all, it will be added to the document with the constraint's value.

the `default`-constraint should be processed after coercing.

example:

```python
>>> schema = {'amount': {'default': 1}}
>>> document = {'amount': None}
>>> v.validated(document, schema)
{'amount': 1}
>>> document = {'model': 'The Robots'}
>>> v.normalized(document, schema)
{'model': 'The Robots', 'amount': 1}
```

nicolaiarocci:
Now that we also transform the incoming document, I guess that stuff like this is welcome. Actually, we implemented a `default` rule in Eve so I know it is useful :) :+1:

funkyfuture:
Is there a difference of Eve's rule-design to my proposal?

nicolaiarocci:
Yes. Of course Eve does not have or need `normalized` since input documents are stored to the backend (with defaults added as needed).

Issue #168, opened 2015-11-04
Bug: schmema validation fails when validating a sub-dict


funkyfuture:
this is a bug i discovered, but i can live with it for a while. anyone feel free to investigate and fix it.

haven't looked much into it, it seems that it affects any `dict`. but i wonder why this isn't covered by tests yet. `allowed` instead of `oneof_regex` behaves alike.

``` python
    def test_of_in_sequence_schema_and_mapping(self):
        schema = {'list': {'schema': {'oneof_regex': ['[a-z]*', '[A-Z]*']}}}
        document = {'list': {}}  # doesn't fail with {'list': []}
        self.assertSuccess(document, schema)
```

```
_________ TestValidation.test_of_in_sequence_schema_and_empty_mapping __________

self = <cerberus.tests.tests.TestValidation testMethod=test_of_in_sequence_schema_and_empty_mapping>

    def test_of_in_sequence_schema_and_empty_mapping(self):
        schema = {'list': {'schema': {'oneof_regex': ['[a-z*]', '[A-Z]*']}}}
        document = {'list': {}}
>       self.assertSuccess(document, schema)

devel/python/cerberus/cerberus/tests/tests.py:1164: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
devel/python/cerberus/cerberus/tests/__init__.py:160: in assertSuccess
    self.assertTrue(validator(document, schema, update),
devel/python/cerberus/cerberus/cerberus.py:552: in validate
    self.__prepare_document(document, normalize)
devel/python/cerberus/cerberus/cerberus.py:605: in __prepare_document
    self.schema)
devel/python/cerberus/cerberus/cerberus.py:422: in _normalize_mapping
    self._normalize_containers(mapping, schema)
devel/python/cerberus/cerberus/cerberus.py:451: in _normalize_containers
    self._normalize_mapping_per_schema(field, mapping, schema)
devel/python/cerberus/cerberus/cerberus.py:490: in _normalize_mapping_per_schema
    purge_unknown=purge_unknown)
devel/python/cerberus/cerberus/cerberus.py:279: in __get_child_validator
    child_validator = self.__class__(**child_config)
devel/python/cerberus/cerberus/cerberus.py:208: in __init__
    self._schema = DefinitionSchema(self, kwargs.get('schema', ()))
[...]

Issue #102, opened 2015-05-26
Proposal: custom coerce-methods


funkyfuture:
i think it's better to include this feature before 0.9 is released. and as there is a reliable pattern for types, it wouldn't be much effort i guess.

one could also add an example to the docs and tests that illustrates a subclassed Validator that makes use of coercing in conjunction with contextual instance-properties.

nicolaiarocci:
Design/usage example? Thanks!

funkyfuture:
```python
class MyValidator(Validator):
    def _corce_none_to_string(self, field, value):
        if value is None:
            self.document[field] = "N/A"

    def _coerce_expand_avatar_path(self, field, value):
        self.document[field] = os.path.join(self.avatar_dir, value)

schema = {'avatar': {'coerce': 'expand_avatar_path', 'type': 'accessible_image_file'},
          'phone': {'coerce': 'none_to_string'}}
```

nicolaiarocci:
:+1:

nicolaiarocci:
[Type coercion](https://cerberus.readthedocs.org/en/latest/#type-coercion) allows this already, as you can pass any callable to the `coerce` method. You can also confine coercion within a custom validator:

    class MyValidator(Validator):
        def __init__(self, *args, **kwargs):
            super(MyValidator, self).__init__(*args, **kwargs)
            self.avatar_dir = '~/avatars'

       def _validate_expand_avatar(self, expand_avatar, field, value):
           self.document[field] = os.path.join(self.avatar_dir, value)

    s = {'a': {'type': 'string', 'expand_avatar': True}

Alternatively you could create a new `avatar` type:

    class MyValidator(Validator):
        def __init__(self, *args, **kwargs):
            super(MyValidator, self).__init__(*args, **kwargs)
            self.avatar_dir = '~/avatars'

       def _validate_type_avatar(self, field, value):
           self.document[field] = os.path.join(self.avatar_dir, value)

    s = {'a': {'type': 'avatar'}}

Do we really need to add yet another way to achieve this?

funkyfuture:
imo yes, as i think that normalization and validation should be clearer segregated in the code. you're proposal is more a workaround where normalization is applied while validation, right?

nicolaiarocci:
Yes, I think you are right.

manugarri:
+1 

One example  would be for those fields in which you want a default value.
Since there is no `default` keyword this could be easily implemented with coerce.

[...]

PR #172, opened 2015-11-19
Add default normalization rule

dnohales:
This is for #131, it adds the `default` normalization rule to add missing fields in the document. Few examples:

```python
>>> schema = {'amount': {'type': 'integer'}, 'kind': {'type': 'string', 'default': 'purchase'}}
>>> v.validated({'amount': 1}, schema)
{'amount': 1, 'kind': 'purchase'}
>>> v.validated({'amount': 1, 'kind': 'other'}, schema)
{'amount': 1, 'kind': 'other'}
>>> v.validated({'amount': 1, 'kind': None}, schema)
{'amount': 1, 'kind': None}
>>> schema = {'thing': {'type': 'dict', 'schema': {'kind': {'type': 'string', 'default': 'purchase'}}
>>> v.validated({'thing': {}}, schema)
{'thing': {'kind': 'purchase'}}
```

### Concerns and thoughts

* If `nullable` is `False`, value is `None` and `default` is present, should we override the field with `default` or throw a validation error?
* If `default` is a callable, should we run the callable and set the callable result? Do we need another option like `default_is_callable` that when is `True` we call the `default` value if is a callable and when is `False` we don't call anything? We also can always call the `default` value if is callable and if the user really want to set a function definition as a default value for the document, she could wrap it with a lambda function or something like that.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -454,6 +454,7 @@ class Validator(object):
         if self.purge_unknown:
             self._normalize_purge_unknown(mapping, schema)
         self._normalize_coerce(mapping, schema)
+        self._normalize_default(mapping, schema)
         self.__normalize_containers(mapping, schema)
         return mapping
 
@@ -563,6 +564,11 @@ class Validator(object):
             new_name = schema[field]['rename_handler'](field)
             mapping[new_name] = mapping[field]
             del mapping[field]
+
+    def _normalize_default(self, mapping, schema):
+        for field in tuple(schema):
+            if 'default' in schema[field] and field not in mapping:
+                mapping[field] = schema[field]['default']
 
     # # Validating
 
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -567,7 +567,10 @@ class Validator(object):
 
     def _normalize_default(self, mapping, schema):
         for field in tuple(schema):
-            if 'default' in schema[field] and field not in mapping:
+            nullable = schema[field].get('nullable', False)
+            if 'default' in schema[field] and \
+                    (field not in mapping or
+                     mapping[field] is None and not nullable):
                 mapping[field] = schema[field]['default']
 
     # # Validating
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -453,8 +453,8 @@ class Validator(object):
         self.__normalize_rename_fields(mapping, schema)
         if self.purge_unknown:
             self._normalize_purge_unknown(mapping, schema)
-        self._normalize_coerce(mapping, schema)
         self._normalize_default(mapping, schema)
+        self._normalize_coerce(mapping, schema)
         self.__normalize_containers(mapping, schema)
         return mapping
 
--- a/cerberus/tests/tests.py
+++ b/cerberus/tests/tests.py
@@ -1319,19 +1319,17 @@ class TestValidation(TestBase):
 
 class TestNormalization(TestBase):
     def test_coerce(self):
-        schema = {
-            'amount': {'coerce': int}
-        }
-        v = Validator(schema)
-        v.validate({'amount': '1'})
-        self.assertEqual(v.document['amount'], 1)
+        schema = {'amount': {'coerce': int}}
+        document = {'amount': '1'}
+        expected = {'amount': 1}
+        self.assertNormalizedEqual(document, expected, schema)
 
     def test_coerce_in_subschema(self):
         schema = {'thing': {'type': 'dict',
                             'schema': {'amount': {'coerce': int}}}}
-        v = Validator(schema)
-        self.assertEqual(v.validated({'thing': {'amount': '2'}})
-                                     ['thing']['amount'], 2)  # noqa
+        document = {'thing': {'amount': '2'}}
+        expected = {'thing': {'amount': 2}}
+        self.assertNormalizedEqual(document, expected, schema)
 
     def test_coerce_not_destructive(self):
         schema = {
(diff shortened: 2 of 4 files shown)


funkyfuture:
cool. cool. cool.

please also extend the documentation. your examples above a fine to be added as `.. doctest::`. maybe except the last.

> If nullable is False, value is None and default is present, should we override the field with default or throw a validation error?

i tend to overriding a `None`. since `nullable` can still prevent it, it seems safe to me. 

> If default is a callable, should we run the callable and set the callable result?

what would be a usecase for that?
i can think of it like a method of a custom validator (thus with context access)? (like #102)

cerberus/cerberus.py
```diff
@@ -454,6 +454,7 @@ def __normalize_mapping(self, mapping, schema):
         if self.purge_unknown:
             self._normalize_purge_unknown(mapping, schema)
         self._normalize_coerce(mapping, schema)
+        self._normalize_default(mapping, schema)
```
funkyfuture: though i have only one silly argument for setting defaults before coercing, here it is: it seems more intuitive to normalize in that order: purge unwanted items, fill up missing items, work on items.

but maybe it's be useful to coerce a value to `None` in order to replace it with a default. coercing defaults on the other hand shouldn't make sense.

so propably it's better to keep it like it is. but i'd be interested in other thoughts.

cerberus/tests/tests.py
```diff
@@ -1438,6 +1438,45 @@ def test_coercion_of_sequence_items(self):
         for x in result['a_list']:
             self.assertIsInstance(x, float)
 
+    def test_default_missing(self):
+        schema = {'foo': {'type': 'string'},
+                  'bar': {'type': 'string',
+                          'default': 'bar_value'}}
+
+        self.assertSuccess({'foo': 'foo_value'}, schema)
```
funkyfuture: you may implement a test-method to test the results of `Validator.normalized`. there should be some tests that can use it. and more will grow.

cerberus/cerberus.py
nicolaiarocci: > but maybe it's be useful to coerce a value to None in order to replace it with a default. coercing defaults on the other hand shouldn't make sense.

Agreed.

nicolaiarocci:
> If nullable is False, value is None and default is present, should we override the field with default or throw a validation error?

I'd say that in this case the user explicitly wants a `None` for the field, so a validation error would probably be more appropriate (let the user know that's not allowed). If we replace it with a default value it might seem it has been accepted while it has not (I'm thinking Eve use case for example, where a client sends a payload to the remote REST API and gets it accepted - I'd probably want to know that `None` is not allowed).

> If default is a callable, should we run the callable and set the callable result?

I'm unsure about this. As @funkyfuture says, there are other ways to achieve the same result already. I probably prefer a static default value.

cerberus/cerberus.py
dnohales: > but maybe it's be useful to coerce a value to None in order to replace it with a default.

But as told by @nicolaiarocci and I also think the same, using default to replace `None` seems odd, specially if `None` is a valid value (`nullable` fields for example). But I know that maybe becomes handy, that's why I proposed to only override `None`s with default only for non `nullable` fields, but the @nicolaiarocci argument in the comment below seems valid too.

> coercing defaults on the other hand shouldn't make sense.

It shouldn't but I don't know if it hurts, maybe it's handy and safer.

dnohales:
> > If default is a callable, should we run the callable and set the callable result?

> what would be a usecase for that?

I was particularly thinking in something like:

```
{'a_datetime': {'default': utcnow}}
```

Then if `a_datetime` is not present in the document, you set the `utcnow` result, not the callable.

But yeah, this over-complicate the feature and makes it more unintuitive. I really don't understand which other way you have to accomplish this, can you give an example?

dnohales:
Updates: added documentation and `assertNormalizedEqual` method.

nicolaiarocci:
Very nice. However this does not seem to work with lists of dictionaries. 

    schema = {
        'mylist': {
            'type': 'list', 
            'schema': {
                'type': 'dict', 
                'schema': {
                    'customer': {'type': 'boolean', 'default': False}, 
                    'name': {'type': 'string'}
                }
           }
        }
    }

Blocks with: `SchemaError: unknown rule 'customer' for field 'customer'`

funkyfuture:
> Then if a_datetime is not present in the document, you set the utcnow result, not the callable.

> But yeah, this over-complicate the feature and makes it more unintuitive. I really don't understand which other way you have to accomplish this, can you give an example?

i don't think this is over-complicated. and i think that's an interesting use-case, so this may come up at some point anyway.
there is no other way to accomplish that now. i didn't mean that there exists one, but that it should be done like custom validation methods. that i could imagine yesterday as use-case, as it would allow context-sensitive defaults.

> If we replace it with a default value it might seem it has been accepted while it has not ... I'd probably want to know that None is not allowed

what about the following:
field value is `None`, `nullable=True` -> `default` is ignored
field value is `None`, `nullable=False` -> field value is set to `default` value, an 'error' is emitted

in conjunction with an ability to ignore specific errors that'd allow flexible usage.

> However this does not seem to work with lists of dictionaries.

i think that may be #168.

nicolaiarocci:
Let's do it this way:

> field value is `None`, `nullable=True` -> `default` is ignored
> field value is `None`, `nullable=False` -> an 'error' is emitted

nicolaiarocci:
For the record, I have a `default` implementation in Eve and I am looking forward to replace that one with this native Cerberus implementation.

funkyfuture:
@nicolaiarocci and Eve's `default` wouldn't touch a `None`, so my proposal would break it?

@dnohales concerning the non-static defaults: that could be achieved if coercing happened after setting defaults; so `default` would ensure that a field is present in a document and could be altered dynamically with a custom coerce method.

i'm curious about this question: are dynamic defaults still in the realm of normalization or is it generation of data? and thus, does it belong into a library with a focus on validation?

nicolaiarocci:
Errors should be show stoppers in my opinion. Especially so when reported by a validation tool. Furthermore, `nullable = False` is a very explicit and clear rule. 

In Eve when `Nullable=False` we return an error and don't process the payload. If `Nullable=True` we accept the null. In both cases we ignore the `default` rule, which only kicks in when field is *missing*.

dnohales:
So, I updated the code. Let me know if you agree with the changes (specially the latest commit), since I see some discrepancies about the compatibility with Eve.

dnohales:
But yeah, I agree with @nicolaiarocci regarding the nullable matter.

nicolaiarocci:
Thanks!

PR #179, opened 2015-12-11
several changes

funkyfuture:
these are some commits that emerged while working towards the XMLErrorHandler.
for some reason i want to drop them.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -475,6 +475,8 @@ class Validator(object):
     # TODO retrieve and store errors from child-validators
     def __normalize_containers(self, mapping, schema):
         for field in mapping:
+            if field not in schema:
+                continue
             if isinstance(mapping[field], Mapping):
                 if 'propertyschema' in schema[field]:
                     self.__normalize_mapping_per_propertyschema(
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -361,14 +361,6 @@ class Validator(object):
             self.__config['error_handler'] = value
 
     @property
-    def _all_errors(self):
-        """
-        :return: A list of all errors, including child errors.
-        """
-        # TODO
-        pass
-
-    @property
     def errors(self):
         """
         Returns the errors of the last processing formatted by the handler that
@@ -966,7 +958,7 @@ class Validator(object):
 
     def _validate_type(self, data_type, field, value):
         def call_type_validation(_type, value):
-            # TODO refactor to a less complex code as submitting an error is now unified and can be triggered here  # noqa
+            # TODO refactor to a less complex code on next major release
             # validator = getattr(self, "_validate_type_" + _type)
             # return validator(field, value)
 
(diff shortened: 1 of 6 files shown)


nicolaiarocci:
So is this a WIP?

funkyfuture:
well, rather not.
three of the commits are needed by the xml-thing, providing general mechanisms. the rest are independent from that. fixing or straightening things.

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/cerberus/commit/4c63700d7f4a1b9d85ee4eb77544833403c5de76

FYI, I don't think it is interesting to add XML/JSON/etc. emission to Cerberus. There are other tools that can do that pretty well so we would end up adding unnecessary features, as I prefer to keep the features set focused on validation (you could also choose to release an external xml-emission tool, we could then mention it in the docs).

I also feel confident that the feature-set is pretty mature, so much that I am actually considering a v1.0 January release.

funkyfuture:
can you point me to a xml-emission tool you think of? next and i guess last would be to en- and decode errors properly. and i'm not keen on reinventing the wheel. though it doesn't seem to be too much of an effort.
yes, i think it's a good idea to keep things focussed. so, i can think of a cerberus-extensions library for various error handlers and additional rules.
thus, i will prioritize to review the `BaseErrorHandler` and `Basic...` to have a stable API. and have a concept-proof for it.

when i reiterated over the type validation i had the strong urge to actually start a 1.0-branch just to see how this could be done much simpler, but compatibility-breaking.
shall i open issues for stuff i think should be done before a 1.0?

nicolaiarocci:
> can you point me to a xml-emission tool you think of?

- https://pypi.python.org/pypi/dicttoxml
- https://stackoverflow.com/questions/1019895/serialize-python-dictionary-to-xml
- https://gist.github.com/huseyinyilmaz/1448723

Withi this being said, I endedup rolling [my own implementation](https://github.com/nicolaiarocci/eve/blob/develop/eve/render.py#L420).

> shall i open issues for stuff i think should be done before a 1.0?

Go for it.

funkyfuture:
thanks for the hints. i did some further research and interestingly there's only `lxml` as maintained library that also has the functionality to convert the resulted xml back to python objects.

i will continue with an own implementation, unless i run into über-effort eventually.
since it is intended as proof-of-concept for the error-handler mechanics i'd like to make it available as reference implementation. shall i just start a repository for the mentioned `cerberus-extensions` and add you with write-permissions?

nicolaiarocci:
> thanks for the hints. i did some further research and interestingly there's only lxml as maintained library that also has the functionality to convert the resulted xml back to python objects.

good ole lxml :+1: 

> since it is intended as proof-of-concept for the error-handler mechanics i'd like to make it available as reference implementation. shall i just start a repository for the mentioned cerberus-extensions and add you with write-permissions?

Sure, but I don't need write permissions I'll just be happy with following the development from afar :smile:
I bet there are other Cerberus "extensions" out there, most likely custom validators. Eventually, it will probably be a good idea to set up a extensions page in the docs, like we did with Eve [extensions page](http://python-eve.org/extensions), so people can add their own tools to it.

PR #196, opened 2016-01-12
Allow None in default normalization rule

dnohales:
Right now, if I try to use this schema:

```
{
    'bar': {
        'type': 'string',
        'nullable': True,
        'default': None
    }
}
```

I get the following exception:

```
cerberus.schema.SchemaError: {'bar': {'default': 'null value not allowed'}}
```

This PR allows us to have `None` as default value.

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -673,6 +673,7 @@ class Validator(object):
             del mapping[field]
 
     def _normalize_default(self, mapping, schema):
+        """ {'nullable': True} """
         for field in tuple(schema):
             nullable = schema[field].get('nullable', False)
             if 'default' in schema[field] and \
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
:+1:

funkyfuture:
Thanks for spotting this. be aware that assertNormalized doesn't validate like the schema seems to intend.

dnohales:
@funkyfuture I don't know if I understood you. The test I added fails as expected without the `nullable` rule.

funkyfuture:
i jzst mean that anything but the default-rule will not be processed in the test. a `self.assertSuccess` should cover these as well.

dnohales:
Is it not worth to check if the document was properly normalized? it doesn't add much extra complexity and can become a useful check.

funkyfuture:
yes, but assertNormalized  doesn't do so. assertSuccess does.

dnohales:
`assertSuccess` only check if the validation was successful, `assertNormalized` also checks that plus compare the normalized document with the document passed in the `expected` parameter.

funkyfuture:
You're right. sorry, I'm knocked out by some virus.

dnohales:
Hehe, no problem!

nicolaiarocci:
Thanks!

PR #198, opened 2016-01-29
Add test which shows recursive schema not allowed

carlos-alberto:


carlos-alberto:
Accidentally opened PR

PR #159, opened 2015-10-12
added minlength / maxlength to dictionaries

fin-8:
The `minlength` / `maxlength` rule, formerly only usable with strings
and lists, is now compatible with arbitrary dictionaries.
Tests and custom field in scheme were added as well.

Updated docs accordingly.

Issue #158, opened 2015-10-12
minlength, maxlength to work with dictionaries


fin-8:
When specifying a dictionary with arbitrary keys using `propertyschema` and `valueschema`, the use of `minlength` would be handy to prevent empty dictionaries.

fin-8:
I've written a patch, pending in Pull Request #159

Issue #146, opened 2015-09-13
Why not officially support blob, date data type? And provide a before, after criterion for datetime and date?


MacHu-GWU:
Why not officially support blob, date data type? And provide a before, after criterion for datetime and date? I know it's easy with custom validator, but since it's easy, why not officially support them?

funkyfuture:
what are you referring to as *blob*?

i can also think of other commonly encountered validations, like for networking and posix-stuff.

for such purposes, i think the best would be to add a module with contributed custom validators in dumb classes that can be used as mix-ins for custom validators:

```
class UserValidator(cerberus.Validator, cerberus.contrib.EMail):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)  # don't try this at home if Python < 3
        self.schema = {'name': {'type': 'string', 'regex': '[a-z]{3,12}'},
                       'email': {'type': 'email address'}}

# or set the schema in your business logic:

class UserValidator(cerberus.Validator, cerberus.contrib.EMail):
    pass
```

funkyfuture:
concerning the before and after criteria: what about covering that by `min` and `max`? but that would be from / to criteria.

funkyfuture:
can this issue be closed? any types can be used with `min` and `max` in the master branch.

PR #199, opened 2016-01-31
changes regarding types and comparisons

### LINKED ISSUE #144: Booleans are accepted for type integer
Internally cerberus 0.10 does the following:

    >>> isinstance(1, (int,))
    True
    >>> isinstance(True, (int,))
    True
    >>> isinstance(0, (int,))
    True
    >>> isinstance(False, (int,))
    True

Of course I could map True and False to integers using Value Coercion, but this seems awkward and feels it is not the correct thing to do when using a type check for integers and not explicitly for booleans.

funkyfuture:
addresses #146
closes #144, #158 and #159

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -1002,14 +1002,12 @@ class Validator(object):
 
     def _validate_maxlength(self, max_length, field, value):
         """ {'type': 'integer'} """
-        if isinstance(value, Sequence):
-            if len(value) > max_length:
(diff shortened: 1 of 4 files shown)


nicolaiarocci:
Spectacular.

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/cerberus/commit/63a082449f54fd984560b4ab00f5493a5d22359a

Issue #176, opened 2015-11-24
Recursion issue when using a recursive schema


carlos-alberto:
We are trying to upgrade to the master branch currently, however the functionality we were currently using to define a recursive schema appears to be broken on the latest release

0.8
===

    In [1]: from cerberus import Validator

    In [2]: schema = {'a': {'type': 'integer'}, 'b': {'keyschema': {'type': 'dict', 'schema': None}}}

    In [3]: schema['b']['keyschema']['schema'] = schema

    In [4]: data = {'a': 1, 'b': {'c': {'a': 10}, 'd': {'a': 10000}}}

    In [5]: validator = Validator(schema)

    In [6]: validator.validate(data)
    Out[6]: True

master
======

    In [1]: from cerberus import Validator

    In [2]: schema = {'a': {'type': 'integer'}, 'b': {'valueschema': {'type': 'dict', 'schema': None}}}

    In [3]: schema['b']['valueschema']['schema'] = schema

    In [4]: validator = Validator(schema)
    ...
    .virtualenvs/connect/lib/python3.4/site-packages/cerberus/cerberus.py in update_to_valueschema(constraints)
       1275                                          "Use 'valueschema' instead.")
       1276         for key, value in constraints.items():
    -> 1277             constraints[key] = update_to_valueschema(value)
       1278         return constraints
       1279

    .virtualenvs/connect/lib/python3.4/site-packages/cerberus/cerberus.py in update_to_valueschema(constraints)
       1275                                          "Use 'valueschema' instead.")
       1276         for key, value in constraints.items():
    -> 1277             constraints[key] = update_to_valueschema(value)
       1278         return constraints
       1279

    .virtualenvs/connect/lib/python3.4/site-packages/cerberus/cerberus.py in update_to_valueschema(constraints)
       1275                                          "Use 'valueschema' instead.")
       1276         for key, value in constraints.items():
    -> 1277             constraints[key] = update_to_valueschema(value)
       1278         return constraints
       1279

    .virtualenvs/connect/lib/python3.4/site-packages/cerberus/cerberus.py in update_to_valueschema(constraints)
       1275                                          "Use 'valueschema' instead.")
       1276         for key, value in constraints.items():
    -> 1277             constraints[key] = update_to_valueschema(value)
       1278         return constraints
       1279

    .virtualenvs/connect/lib/python3.4/site-packages/cerberus/cerberus.py in update_to_valueschema(constraints)
       1267     # TODO remove on next major release
[...]

PR #201, opened 2016-02-04
Schema validation

funkyfuture:
the main aspect of this pr is the usage of child-validators for schema-validation in order to make that non-blocking.
along with it are changes that are necessary to make that work and smaller fixes.

i also wanted to implement a cache that tracks transformed schemas for optimization and to tackle #176, but since mappings are not hashable that requires some caution, and thus time, and i didn't want to hold back my progress so far.


nicolaiarocci:
Looks good.

nicolaiarocci:
Rebased and merged with https://github.com/nicolaiarocci/cerberus/commit/dc2f74a028318fb79958d022122fa0b7a9906830

Thanks.

PR #207, opened 2016-02-28
Added Support for `binary` Type

seglberg:
The binary type is similar to string except for Python 2 it only accepts `str` and for Python 3 it accepts `bytes` or `bytearray`.

---
Example in Python 3:
```python
In [2]: v = Validator({'mybinary': {'type': 'binary'}})

In [3]: v.validate({'mybinary': b'valid'})
Out[3]: True

In [4]: v.validate({'mybinary': bytearray(b'valid')})
Out[4]: True

In [5]: v.validate({'mybinary': 'invalid'})
Out[5]: False
```

[Review on Reviewable](https://reviewable.io/reviews/nicolaiarocci/cerberus/207)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,3 +36,4 @@ Patches and Suggestions
 - Tobias Betz
 - Trong Hieu HA
 - calve
+- Matthew Ellison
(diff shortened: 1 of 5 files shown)


nicolaiarocci:
Rebased and merged with https://github.com/nicolaiarocci/cerberus/commit/f6ae3b529bbd9e47bb86e3ff4be8bd253d22e0d0

Thanks!

PR #224, opened 2016-04-06
Improve Warning on `propertyschema` Normalization

### LINKED ISSUE #222: Property Schema Now Causing Warning
This is probably more of a question, than an issue. However, did the correct way to use `propertyschema` change recently on master (aka upcoming 0.10/1.0)? I now get a UserWarning, see below:

```python
schema = {
    "users": {
            "type": "dict",
            "required": True,
            "propertyschema": {
                "type": "string"
            }
      }
}

v = Validator(schema)
v.validate({"users": {"test_user": "user_information"}})

cerberus/cerberus.py:594: UserWarning: Normalizing keys of users: test_user already exists, its value is replaced.
```

Not sure if this is an issue or if there is a change in behavior that I need to adopt.

seglberg:
If the propertyschema ("key") did not change during normalization, do not emit a warning about it being replaced.

Closes #222

---
This change is [Reviewable](https://reviewable.io/reviews/nicolaiarocci/cerberus/224)

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -587,6 +587,8 @@ class Validator(object):
             self._drop_nodes_from_errorpaths(validator._errors, [], [2, 4])
             self._error(validator._errors)
         for k in result:
+            if k == result[k]:
+                continue
             if result[k] in mapping[field]:
                 warn("Normalizing keys of {path}: {key} already exists, "
                      "its value is replaced."


funkyfuture:
if it works, it lgtm.

i'd used a generator comprehension, but i think i'm compulsive concerning comprehensions.

seglberg:
Indeed, it's easy to throw generators everywhere. Actually I'm surprised I didn't in this scenario...

Anyways, I locally ran a few of my projects with my updated fork and the warnings have now disappeared.

nicolaiarocci:
Thanks!

Issue #225, opened 2016-04-15
Cerberus validator


txomon:
I have been using cerberus for a while, and I actually miss a feature, would it be possible to have a cerberus validator? like a validator to validate schemas?

I often find myself creating really complex schemas, and spotting one missing level, or a keyword, and having a way to validate my schemas would be really welcome.

I will probably end up doing it myself eventually, but just in case I wanted to post the idea.

funkyfuture:
schema validation is actually implemented for a while, but has has been pimped since the last release. if you need an impression of the latest development, i can open a pr that includes the results of #217.

if you're currently using the master branch and no `SchemaError` is raised due to a broken schema, consider it as a bug.

txomon:
I was thinking on using cerberus validate() to validate the schema, having
the same output as the normal cerberus validation. Now the validation is
quite cryptic and doesn't give me as many insights as validate(). I was
thinking on creating an schema that validates schemas.

txomon:
Just in case, I was referring to the output of:

```
from cerberus import Validator

sample_schema = {
    'id': {
        'type': 'string',
        'nullable': False,
        'required': False,
    },
    'kind': {
        'type': 'dict',
        'anyof': [
            {
                'schema': {
                    'type': {
                        'type': 'string',
                        'allowed': ['kind1'],
                        'required': True,
                        'nullable': False,
                    },
                    'data11': {
                        'type': 'integer',
                    },
                },
            },
            {
                'schema': {
                    'type': {
                        'type': 'string',
                        'allowed': ['kind2'],
                        'required': True,
                        'nullable': False,
                    },
                    'data21': {
                        'type': 'string',
                    },
                },
            },
            {
                'schema': {
                    'type': {
                        'type': 'string',
                        'allowed': ['kind3'],
                        'required': True,
                        'nullable': False,
                    },
                    'data31': {
                        'type': 'datetime',
                    },
                },
            },
        ],
    },
    'temp': {
        'type': 'list',
        'anyof': [
            {
                'schema': {
                    'type': {
                        'type': 'string',
                        'required': True,
                        'nullable': False,
                        'allowed': ['temp1'],
                    },
                },
            },
            {
                'schema': {
                    'type': {
                        'type': 'string',
                        'required': True,
                        'nullable': False,
                        'allowed': ['temp2'],
                    },
                },
            },
            {
                'schema': {
                    'type': {
                        'type': 'string',
                        'required': True,
                        'nullable': False,
[...]

PR #226, opened 2016-04-19
Creating testcase featuring strange behaviour

txomon:
Following #225, I found a validator issue that I don't fully understand.

The main idea of the new test case file is to create end to end test cases. Feature use cases that have been broken in some point in the history, so that no regression happens.

### DIFF
--- a/cerberus/tests/test_usecases.py
+++ b/cerberus/tests/test_usecases.py
@@ -1,0 +1,44 @@
+from unittest import TestCase
+
+from cerberus import Validator
+
+
+class TestComplexCases(TestCase):
+    def test_nested_list_choice_dicts(self):
+        test_schema = {
+            'temp': {
+                'type': 'list',
+                'anyof': [
+                    {
+                        'schema': {
+                            'type': {
+                                'type': 'string',
+                                'required': True,
+                                'nullable': False,
+                                'allowed': ['tmp1'],
+                            },
+                            'args': {
+                                'type': 'list',
+                                'schema': {
+                                    'type': 'string'
+                                }
+                            }
+                        },
+                    },
+                ]
+            }
+        }
+        validator = Validator(schema=test_schema)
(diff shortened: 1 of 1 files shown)


txomon:
I must say that I have no idea on how to fix it. My guess is that anyof together with schema doesn't work quite well.

Just in case, check also the schema correctness...

funkyfuture:
i agree, the should be more complex testcases. though i recently removed one. but i think it's better put on hold until a `1.0`.

however when identifying and solving bugs i prefer the least complex setup.

> Just in case, check also the schema correctness...

well, it makes not much sense as you provide only one variant for `anyof`. is this intended?

to get an idea of what may be wrong, the output of this test would be useful. i'm currently focussed on something else when i have the time to work on cerberus, hence i won't look into it too soon..

please have a look at `TestBase`.

txomon:
You have the output in the CI server, will copy it ASAP.

Probably I didn't express myself correctly, I wanted to add regression test
cases, which sometimes are complex.

In this case I reduced the example to a bare minimum to expose the bug.

Looks like there is a recursion bug on constraint checking, but I really
can't tell

funkyfuture:
nope, the interesting part is here: `AttributeError: 'Validator' object has no attribute '_validate_type_required'`

which means that cerberus wants to validate a type that isn't implemented. this is caused by the ambiguity of `schema`. it /should/ work when you define a `'type': 'list'` beside the first `schema`-rule.

see #168 and 7dc632866a2ea6956dc65f42d03e3ab45c295605.

if you find anything to improve the situation, it's highly appreciated.

funkyfuture:
@txomon is this still stomething that needs to be addressed? after rebasing on top of master's HEAD.

txomon:
Sorry, I am unable to book time for this :/ I will have a look on it if I step again on it.

PR #227, opened 2016-04-21
moving forward

funkyfuture:
initially i wanted to use cerberus' normalization capabilities to expand a validation schema. eventually it turned out that would require extra complexity (e.g. allowing `rename_handler` for unknown fields). but the aim was to reduce complexity. and it's hardly much less in the end, but well, some polishing is achieved. while working on it, i removed some of backward compatibility since i saw no sense in refactoring that pieces anymore. so, it also has a first shot on `UPGRADING.rst`.


UPGRADING.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Deprecations
+------------
+
+items (for mappings)
+~~~~~~~~~~~~~~~~~~~~
+
+The usage of the ``items``-rule is restricted to sequences.
+If you still had schemas that used that rule to validate :term:`mappings`,
+just rename these instances to ``dict``.
```
funkyfuture: should be ``schema``, not ``dict``

nicolaiarocci:
Thanks!

PR #228, opened 2016-04-24
Adds capability to use references in schemas

funkyfuture:
it turned out to implement this quiet straight-forward.
there are yet no docs, i will amend them if this is considered okay.

### DIFF
--- a/cerberus/__init__.py
+++ b/cerberus/__init__.py
@@ -9,13 +9,16 @@
 """
 
 from .cerberus import Validator, DocumentError
-from .schema import SchemaError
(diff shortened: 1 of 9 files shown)


nicolaiarocci:
Very nice. 👍

funkyfuture:
rebased and docs amended.

i wonder if the section *Allowing The Unknown* in `usage.rst` should also be moved to the new `schema.rst`.

nicolaiarocci:
Hmm unsure about that. It is true that it deals with schemas, but it is probably nice to have it on the usage page right away. Maybe just add a reference in schema.rst? Or vice-versa.

Issue #187, opened 2015-12-23
Proposal: Change return value of specific type checks


funkyfuture:
with the introduction of the possibility to allow multiple types, the code became hackish where
type checks are delegated. the addition of error handlers even worsened it. however they also make it possible to refactor the code, that a specific type test would return a boolean.
as this breaks code with custom type-checks i think the opportunity of a 1.0 release should be taken
to bring that essential code into a reasonable form. the refactoring effort for client code seems
affordable.

nicolaiarocci:
Actually I am very concerned about breaking changes. Just as an example I will probably have to delay 1.0 adoption in Eve just because it would end up breaking Eve custom validators (Eve being used on a lot of production systems). If I had the time I would go back and look at the breaking changes and see how/if we can smooth the transition process by going the deprecation path (like we did in different areas, supporting/deprecating legacy stuff)

funkyfuture:
i think the difficulty arises from these points:
- if a field is validated against multiple legit types, each type's validation method is called
- type validation methods do each submit an (identical) error if failing
  - which can't and shouldn't be revoked because who knows what an error-handler did when its `emit`-method was called
- minor: it'd be easier to figure out whether to signal a validation abortion by just checking a bool value rather than measuring if an error was submitted

so, the current use of child-validators is necessary to not log invalid errors. but it also uses a lot of overhead.

i just had the idea that maybe the `_error` method could be tricked. hm, that could work as a workaround for a transition release. but i would suggest do make that still a beta-release and drop all obsolete stuff (btw, there seems to be a discussion to drop Py2.6 support in py.test) in a first final release.

nicolaiarocci:
Addessed by #231

PR #231, opened 2016-05-03
out with the old

funkyfuture:
here are some changes that shouldn't leave much to do for a new release. all deprecated parts are removed.

the last commit implements what i proposed in #187 and the changing of the methods went very well. it also seems to have a mentional impact on performance. the runtime of the tests reduced from ~800ms to ~600ms on my system.


nicolaiarocci:
Merged with https://github.com/nicolaiarocci/cerberus/commit/27925cd4dd25cad25dc97e030cf47fd924a638b5

Issue #210, opened 2016-03-05
Why error value can be string and list of strings?


andreymal:
It is uncomfortable for me. I have to write extra type checking.

This is OK (except #209, but it is not about that):
```python
>>> v = Validator({'name': {'type': 'string', 'minlength': 6, 'regex': '[A-z]+'}})
>>> v.validate({'name': '©'})
False
>>> for field, value in v.errors.items():
...  print(field)
...  for error in value:
...   print(' -', error)
... 
name
 - max length is 6
 - value does not match regex '[A-z]+'
```

But:
```python
>>> v.validate({'name': 'abc'})
False
>>> for field, value in v.errors.items():
...  print(field)
...  for error in value:
...   print(' -', error)
... 
name
 - m
 - a
 - x
 -  
 - l
 - e
 - n
 - g
 - t
 - h
 -  
 - i
 - s
 -  
 - 6
```

WAT

I don't wish write `if isinstance(value, str)`!

nicolaiarocci:
This happens on the master branch, right? Should not be on stable (v0.9.2).

funkyfuture:
nah, that's the behaviour how errors are stored in `Validator.errors` also before `0.10`. but it's also implemented in the `BasicErrorHandler` for backward compatibility.

@andreymal there's also a lower-level access to errros in the master branch: http://docs.python-cerberus.org/en/latest/errors.html

andreymal:
Lower-level access is too low :)

>for backward compatibility.

I only can deal with it and write a custom errorhandler?

funkyfuture:
with a new major relase coming, this behavior could be changed. there shouldn't be an impact on client code unless it could handle single errors on a field anyway.

nicolaiarocci:
👍

Issue #237, opened 2016-05-11
Problem with validation of keys in dict type 


tadeu:
I'm trying to validate a dict type where the keys may be any string. This is a simple example:

```python
schema = {
    'test': {
        'type': 'dict',
        'keyschema': {'type': 'string'},
        'valueschema': {'type': 'integer'},
    },
}

v = Validator(schema)
v.validate({'test': {'a': 1, 'b': 2}})
print v.errors
```

Output:
```
WARNING cerberus: `keyschema` is deprecated, use `valueschema` instead
{'test': [{'a': 'must be of string type'}, {'b': 'must be of string type'}]}
```

(Using Cerberus 0.9.2)
What is wrong? Thanks!

funkyfuture:
in this version `keyschema` is an alias of `valueschema` for historical reasons.

either use `propertyschema` instead or use the current `master`-branch's `HEAD`.

and always rtfm, especially when confronted with an obvious warning.

tadeu:
Ok, I've "RTFM", but it is confusing. So let me rephrase the problem.
How can I validate documents like this:

```
{'test': {'a': 1, 'b': 2, 'c': 3}}
```

Where `test` is required, and it is a dict composed of any number of key/values, but the keys must follow a schema (in this case, to be a simple string), and the values must follow a different schema (in this case, to be an integer).

funkyfuture:
did you read the `stable` or the `latest` docs? which version do you intend to use? what is confusing in particular?

tadeu:
In both, but I was trying to accomplish this with the `stable` version.

For example, this is confusing in both documentations:

> ### `keyschema`
> This is the counterpart to `valueschema` that validates the _keys_ of a dict. **For historical reasons it is not named `keyschema`.**

Also it is not clear in the stable documentation how to accomplish what I want, but at least it has got this not in the `latest` documentation:

> ### Note
> To validate arbitrary keys of a mapping, see `keyschema`, resp. `valueschema` for validating arbitrary values of a mapping.

Well... guess I'll have to try working on `master`...

funkyfuture:
lol, that was rather a search & destroy and not replace when i recently updated the docs. thanks for pointing that out.

your example should work as expected with the current master.

Am 11. Mai 2016 17:41:52 MESZ, schrieb Tadeu Manoel <notifications@github.com>:

Issue #235, opened 2016-05-10
overwriting mandatory_validations in custom Validator fails


stijnvanhoey:
When adapting the mandatory_validations, the validation of the added mandatory validation fails:

```python
from cerberus import Validator

class MyValidator(Validator):
    mandatory_validations = ('empty', 'nullable')

val = MyValidator({'user':{'allowed': ['Myriam', 'Floris']}})
document = {'user': ''}
val.validate(document)
print(val.errors)
```

Produces the following error:

```python
 ...
  File "/cerberus/cerberus/cerberus.py", line 902, in _validate_empty
    self._error(field, errors.EMPTY_NOT_ALLOWED)

  File "/cerberus/cerberus/cerberus.py", line 242, in _error
    constraint = field_definitions[rule]

KeyError: 'empty'
```
I would expect that this kind of adaptation by a custom class should be possible without providing this error?

funkyfuture:
your usecase fails because you declare a rule as mandatory and then not specify that rule in your schema.
a fix should therefore raise an error in such cases while schema validation.
the ’nullable’ rule is the only one that has a default constraint. I guess the reasoning behind is that every field that you declare in a schema is in almost all cases expected to have a value (which is to be validated).
so, imo #235 is the wrong approach, rather a hotfix than a solution.

stijnvanhoey:
In interpreted this differently as I was expecting that `mandatory_validations` are validations that will always be executed with a default value (cfr. nullable is False) even when not added to the schema. However, it can be added to the schema to overwrite the default behavior (nullable as True). `Nullable` is in cerberus mandatory, as in: you do not have to add it to the schema to execute it.

nicolaiarocci:
> your usecase fails because you declare a rule as mandatory and then not specify that rule in your schema. a fix should therefore raise an error in such cases while schema validation.

Agreed. Maybe we could update the API docs with this, so people don't get confused anymore.

funkyfuture:
It also should be simple to let the metaclass add a required rule for mandatory rules into the schema validation schema.

funkyfuture:
@stijnvanhoey i added a fix with 0031381. you may submit a test for this derived from you use-case.

PR #239, opened 2016-05-12
Streamlining BasicErrorHandler

funkyfuture:
here's a solution to #210.


funkyfuture:
also addresses https://github.com/nicolaiarocci/cerberus/issues/237#issuecomment-218499564

nicolaiarocci:
It also addresses #235

Issue #72, opened 2015-03-11
update mode in subdocuments


ecarreras:
How can I make that subdocuments work in update mode (aka no validate required fields)

ecarreras:
By now I've a subclass of validator which propagates `update` in `validate` method. Is there another way?

nicolaiarocci:
Ah right, should be fixed now. Thanks!

PR #244, opened 2016-06-13
Fixed list of subdict with update=True

JonathanHuot:
validate(..update=True) does not work when required fields are in a *list* of subdicts. This issue is different from https://github.com/nicolaiarocci/cerberus/issues/72 because the code is different for subdicts from list of subdicts.

Steps to reproduce (tested HEAD as well, and submitted PR against 1.0pre):

```
>>> import cerberus
>>> cerberus.__version__
'0.9.2'
>>> schema = {'items': {'type': 'list', 'schema': {'type': 'dict', 'schema': {'name': {'required': True, 'type': 'string'}, 'age': {'type': 'integer'}}}}}
>>> v = Validator(schema)
>>> v.validate({'items': [{'age': 10}]})
False
>>> v.errors
{'items': {0: {'name': 'required field'}}}
>>> v.validate({'items': [{'age': 10}]}, update=True)
False
>>> v.errors
{'items': {0: {'name': 'required field'}}}
```

### DIFF
--- a/cerberus/cerberus.py
+++ b/cerberus/cerberus.py
@@ -1111,7 +1111,7 @@ class Validator(object):
         validator = self._get_child_validator(
             document_crumb=field, schema_crumb=(field, 'schema'),
             schema=schema, allow_unknown=self.allow_unknown)
-        validator(dict(((i, v) for i, v in enumerate(value))), normalize=False)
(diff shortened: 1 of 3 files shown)


nicolaiarocci:
Thanks!

PR #247, opened 2016-07-14
Fixes #241 and other

### LINKED ISSUE #241: dependencies cannot be used inside of oneof ?
Using cerberus version 1.0rc

I'm trying to structure this, so that if `a` is `1` then `b` is required, else if `a` is `2` then `c` is required. Is there something wrong with my usage here, or is this a bug?

```
>>> schema = {
...     "a": {
...         "type": "integer",
...         "oneof": [
...             {
...                 "allowed": [
...                     1
...                 ],
...                 "dependencies": [
...                     "b"
...                 ]
...             },
...             {
...                 "allowed": [
...                     2
...                 ],
...                 "dependencies": [
...                     "c"
...                 ]
...             }
...         ]
...     },
...     "b": {
...         "type": "string"
...     },
...     "c": {
...         "type": "string"
...     }
... }
>>> v.validate({"a": 1, "b": "123"}, schema)
False
>>> v.errors
{'definition 1': [{'a': ['unallowed value 1', "field 'c' is required"]}], 'a': [{'oneof': ['none or more than one rule validate']}, "field 'b' is required", 'unallowed value 1', "field 'c' is required"], 'definition 0': [{'a': ["field 'b' is required"]}]}
>>> v.validate({"a": 2, "c": "123"}, schema)
False
>>> v.errors
{'definition 1': [{'a': ["field 'c' is required"]}], 'a': [{'oneof': ['none or more than one rule validate']}, 'unallowed value 2', "field 'b' is required", "field 'c' is required"], 'definition 0': [{'a': ['unallowed value 2', "field 'b' is required"]}]}
```

funkyfuture:
here's a fix for #241 included. @gwainer can you check that it also covers your issue?

i hope i didn't miss anything, the solution was not as straight-forward as finding the problem. there's been a blind spot concerning the affects of other rules so far.

why did i end up renaming the `cerberus` module? well,

1. i wanted to run the tests on Pyston and thus needed an invokable script. (the result is surprisingly only one fail that may be a bug in Pyston's exception system.) unfortunately it doesn't work with pytest yet.
2. i noticed at this point that the imports all don't follow PEP8 which favor absolute imports and fixed it.
3. suddenly it looked javaish (`from cerberus.cerberus import x`) and was ambigous (`from cerberus.import x`), hence the renaming

there's no rush to merge as i will have a look at the incorrect error representation from the initial post.

### DIFF
--- a/cerberus/__init__.py
+++ b/cerberus/__init__.py
@@ -8,8 +8,9 @@
 
 """
 
-from .cerberus import Validator, DocumentError
-from .schema import rules_set_registry, schema_registry, Registry, SchemaError
+from cerberus.validator import Validator, DocumentError
+from cerberus.schema import (rules_set_registry, schema_registry, Registry,
+                             SchemaError)
 
 
 __version__ = "1.0rc"
--- a/cerberus/__init__.py
+++ b/cerberus/__init__.py
@@ -8,6 +8,8 @@
 
 """
 
+from __future__ import absolute_import
+
 from cerberus.validator import Validator, DocumentError
 from cerberus.schema import (rules_set_registry, schema_registry, Registry,
                              SchemaError)
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -3,7 +3,8 @@
 from __future__ import absolute_import
 
 from collections import defaultdict, namedtuple, MutableMapping
-from copy import copy
+from copy import copy, deepcopy
+from pprint import pformat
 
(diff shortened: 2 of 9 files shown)


gwainer:
Just tested it @funkyfuture, and it solves the problem I mentioned in #241. Nice work! Thanks.

funkyfuture:
it's ready to merge

nicolaiarocci:
Excellent.

Issue #250, opened 2016-07-26
Unexpected TypeError when using allow_unknown in a schema defining a list of dicts


daviskirk:
I'm not sure if I understand the 'allow_unknown' rule correctly:

I can do this

```python
import cerberus
schema = {
    'test': {
        'schema': {'a': {'type': 'string'}}, 'type': 'dict', 'allow_unknown':True
    }
}

validator = cerberus.Validator(schema)
validator.validate({'test': {'unknown_key': 'unknown_value'}})
validator.validate({'test': {}})
```
and both validations return `True`.

But if I nest the dictionary inside a list and try to do the same:

```python
schema = {
    'test': {
        'type': 'list',
        'schema': {
            'type': 'dict',
            'allow_unknown':True,
            'schema': {'a': {'type': 'string'}}
        }
    }
}
validator = cerberus.Validator(schema)
validator.validate({'test': [{'unknown_key': 'unknown_value'}]})
validator.validate({'test': {}})
```
unexpected things (for me at least) happen.
The first validation results in `True` as expected.  For the second however, I would have expected it to return `False` and indicate an error (because of the dict was used instead of the list) within the `errors` attribute.  Instead, cerberus throws a `TypeError` while calling __getitem__ on the schema using the 'allow_unknown` value `True` which of course fails. 

Am I using the rule incorrectly?  I am using cerberus 1.0rc0 pulled directly from github.

```pdb
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-44-bead38c40317> in <module>()
----> 1 validator.validate({'test': {}})

/pythonenv/lib/python3.5/site-packages/cerberus/validator.py in validate(self, document, schema, update, normalize)
    730         if normalize:
    731             self.__normalize_mapping(self.document, self.schema)
--> 732 
    733         for field in self.document:
    734             if self.ignore_none_values and self.document[field] is None:

/pythonenv/lib/python3.5/site-packages/cerberus/validator.py in __normalize_mapping(self, mapping, schema)
    497         self.__normalize_default_fields(mapping, schema)
    498         self._normalize_coerce(mapping, schema)
--> 499         self.__normalize_containers(mapping, schema)
    500         return mapping
    501 

/pythonenv/lib/python3.5/site-packages/cerberus/validator.py in __normalize_containers(self, mapping, schema)
    552                 if set(schema[field]) & set(('allow_unknown', 'purge_unknown',
    553                                              'schema')):
[...]

PR #255, opened 2016-08-26
fix issue #250

daviskirk:
This is a start at fixing issue #250 
I looked into this and the problem seems to be that by passing a mapping, the normalizer goes through the "mapping normalizer" code path, which leads to all sorts of problems once the boolean value of "allow_unknown" is encountered.  
The "schema" keyword means something different when set in the context of a sequence (in a mapping, the top level of the schema dictionary defines keys of the schema, in a sequence, it defines the top level is equivalent to the second level of a mapping schema), resulting in the exceptions thrown at the normalization stage.
This has probably been discussed before but if not, it might be sensible to use a different keyword for the lists "schema" like "itemschema" or something like that?

### DIFF
--- a/docs/validation-rules.rst
+++ b/docs/validation-rules.rst
@@ -606,13 +606,14 @@ A list of types can be used to allow different values:
 
 .. note::
 
-    While the ``type`` is not required to be set at all, it is not encouraged to
-    leave it unset especially when using more complex rules such as ``schema``.
-    If you decide you still don't want to set an explicit type, rules such as
-    ``schema`` are only applied to values where the rules can actually be used
-    (such as `'dict'` and `'list'`). Also, the the case of ``schema``, cerberus
-    will try to decide depending on what the ``schema`` rule looks like if a
-    `'list'` or a `'dict'` is required.
+    While the ``type`` rule is not required to be set at all, it is not
+    encouraged to leave it unset especially when using more complex rules such
+    as ``schema``. If you decide you still don't want to set an explicit type,
+    rules such as ``schema`` are only applied to values where the rules can
+    actually be used (such as ``dict`` and ``list``). Also, in the case of
+    ``schema``, cerberus will try to decide if a ``list`` or a ``dict`` type
+    rule is more appropriate and infer it depending on what the ``schema`` rule
+    looks like.
 
 .. note::
 
--- a/docs/validation-rules.rst
+++ b/docs/validation-rules.rst
@@ -606,6 +606,16 @@ A list of types can be used to allow different values:
 
 .. note::
 
+    While the ``type`` is not required to be set at all, it is not encouraged to
+    leave it unset especially when using more complex rules such as ``schema``.
+    If you decide you still don't want to set an explicit type, rules such as
+    ``schema`` are only applied to values where the rules can actually be used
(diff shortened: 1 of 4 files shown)


nicolaiarocci:
cc @funkyfuture

nicolaiarocci:
Looks good to me!

funkyfuture:
yeah, the problem with the ambiguity is known.

you should also add a test case where type is not defined in the schema.

here's a solution for that issue in validating: 7dc632866a2ea6956dc65f42d03e3ab45c295605

daviskirk:
I didn't know it was possible to do "implicit" list validation without specifying the type at all depending on how the schema looks.  This seems kind of counter intuitive.  Just for clarification:

As I understand this implicit schema parsing works like this:

```python
schema = {'list': {'schema': {'allowed': ['a','b','c']}}}
v = Validator(schema)
v.validate({'list': ['a', 'b']})  
# True, interprets the schema type of 'list' as a list
v.validate({'list': {'a': 'something'}})  
# error {'list': ['must be of dict type']}, because cerberus sees the 'allowed' in the schema an assumed you meant a list
```

However, I can also do:

```python
v.validate({'list': 4})
# also True, apparently because the schema is only checked if a list or dict is passed?
```

I've restructured the checking with the help of your commit (thanks) @funkyfuture .  And I think I've covered the cases I described above... assume I've understood the mechanics correctly.

To add to the discussion:
Instead of the TypeError check of the list comprehension, I could also go further with the error checking and only raise the _SchemaRuleTypeError if 'allow_unknown' is the problem.  However I don't know if this is more help or harm depending on what other rules you could throw in there that don't have valid mappings as values.

funkyfuture:
on the train I can't really browse through the code and docs. but concerning your last question I think, it's okay as it is. rules are usually very fault tolerant outside their purpose. with that ambiguity and possible complementing rules to schema though, this can't be applied consequently.

so, lgtm. maybe some note to the docs to prevent misunderstanding. you know best. :-)

(and all your assumptions are correct, it think)

daviskirk:
I added a note to the docs, wasn't sure whether to put something under the `schema` rule, but since it really only gets hairy when you don't define a `type` rule I would think that's the better place for it.

docs/validation-rules.rst
```diff
@@ -606,6 +606,16 @@ A list of types can be used to allow different values:
 
 .. note::
 
+    While the ``type`` is not required to be set at all, it is not encouraged to
+    leave it unset especially when using more complex rules such as ``schema``.
+    If you decide you still don't want to set an explicit type, rules such as
+    ``schema`` are only applied to values where the rules can actually be used
+    (such as `'dict'` and `'list'`). Also, the the case of ``schema``, cerberus
```
funkyfuture: please surround type/class names with `` or even link the Python docs

    :class:`dict`

daviskirk: I updated this to make it clearer that ``type`` and ``dict`` refer to the cerberus type rules and not the explicit python classes.

docs/validation-rules.rst
```diff
@@ -606,6 +606,17 @@ A list of types can be used to allow different values:
 
 .. note::
 
+    While the ``type`` rule is not required to be set at all, it is not
```
nicolaiarocci: excellent

Issue #213, opened 2016-03-05
Proposal: completely switch to py.test


funkyfuture:
this tools has been released recently: https://github.com/dropbox/unittest2pytest

if it does its job good, it shouldn't be such an effort to rewrite all tests as they are usable with `pytest`.

i'd be much in favor for it, alone for the heavily reduced char count of the `tests.py`. but certainly after an 1.0 release.

nicolaiarocci:
Interesting.

funkyfuture:
just two cents:

- used it. it's realiable. it's not without straight forward effort anyway. but a good start.
- pytest doesn't run on pyston yet.

nicolaiarocci:
Partially addressed by #259

PR #256, opened 2016-08-26
Fixes #251

### LINKED ISSUE #251: allow_unknown inside *of rule is ignored
When trying to use `allow_unknown` within an `*of` clause, validation fails. 
This is perhaps related to #250 , but as no exception is thrown in this case and the error tree is seemingly successfully built and only ignores the `allow_unknown` flag, I'm opening a separate issue.

```python

# This is the schema that causes the problem
schema1 = {
    'test': {
        'oneof': [
            {
                'type': 'dict', 
                'allow_unknown':True, 
                'schema': {'a': {'type': 'string'}}
            }
        ]
    }
}

# This is a control schema, meaning it is the same as the above, 
# just without wrapping the `oneof` clause around the inner `'dict'`
schema2 = {
    'test': {
        'type': 'dict', 
        'allow_unknown':True, 
        'schema': {'a': {'type': 'string'}}
    }
}

v1 = cerberus.Validator(schema1) 
v1.validate({'test': {'a': 's'}}) #  returns True (as expected) 
v1.validate({'test': {'a': 's', 'b': 'asd'}}) #  returns False (is `allow_unknown` ignored?)
print(v1.errors) #  {'test': [{'oneof': ['none or more than one rule validate', {'oneof definition 0': ['unknown field']}]}]}

v2 = cerberus.Validator(schema2)  
v2.validate({'test': {'a': 's', 'b': 'asd'}}) #  returns True (as expected)
```

daviskirk:
- add test

### DIFF
--- a/cerberus/tests/tests.py
+++ b/cerberus/tests/tests.py
@@ -1292,6 +1292,37 @@ class TestValidation(TestBase):
         self.assertFail({'a': 1, 'c': 'foo'}, schema)
         self.assertFail({'a': 2, 'b': 'bar'}, schema)
 
+    def test_allow_unknown_with_oneof_rules(self):
+        # https://github.com/nicolaiarocci/cerberus/issues/251
+        schema = {
+            'test': {
+                'oneof': [
+                    {
+                        'type': 'dict',
+                        'allow_unknown': True,
+                        'schema': {'known': {'type': 'string'}}
+                    },
+                    {
+                        'type': 'dict',
+                        'schema': {'known': {'type': 'string'}}
+                    },
+                ]
+            }
+        }
+        # check regression and that allow unknown does not cause any different
+        # than expected behaviour for one-of.
+        document = {'test': {'known': 's'}}
+        self.validator(document, schema)
+        _errors = self.validator._errors
+        self.assertEqual(len(_errors), 1)
+        self.assertError('test', ('test', 'oneof'),
+                         errors.ONEOF, schema['test']['oneof'],
+                         v_errors=_errors)
+        self.assertEqual(len(_errors[0].child_errors), 0)
+        # check that allow_unknown is actually applied
+        document = {'test': {'known': 's', 'unknown': 'asd'}}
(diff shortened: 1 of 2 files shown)


daviskirk:
As this wasn't really related to #250 after all I looked at this as well while I was at it. As far as I can see, this is now the intended behaviour. 

I added a test for oneof, assuming the other *of rules should work the same.  Otherwise I can add the others too.... it might be alot of duplicated code that might just result in harder to maintain tests (As a side note I saw #213 and pytest parametrized tests are great cutting down on that sort of stuff).

cerberus/tests/tests.py
```diff
@@ -1580,6 +1580,36 @@ def test_coerce_non_digit_in_sequence(self):
...
+                    },
+                ]
+            }
+        }
+        # check regression and that allow unknown does not cause any different
+        # than expected behaviour for one-of.
+        document = {'test': {'known': 's'}}
+        self.validator(document, schema)
+        _errors = self.validator._errors
+        self.assertEqual(len(_errors), 1)
```
funkyfuture: please also check that there are no child-errors.

cerberus/tests/tests.py
```diff
@@ -1580,6 +1580,36 @@ def test_coerce_non_digit_in_sequence(self):
...
+        self.assertError('test', ('test', 'oneof'),
+                         errors.ONEOF, schema['test']['oneof'],
+                         v_errors=_errors)
+        # check that allow_unknown is actually applied
```
funkyfuture: *of-rules are ignored during normalization. am i missing something? otherwise i see no point in it.

daviskirk: Yes, you're 100% correct... testing normalization makes no sense.

nicolaiarocci:
Merged after rebase: https://github.com/nicolaiarocci/cerberus/commit/78704de458d50db0f38fb614aef1eeb09c34f3de

Thanks!

PR #267, opened 2016-09-12
Consistently use new-style classes

dkellner:
We should use new-style classes wherever possible. There were still four classes not inheriting ```object``` which will lead to confusion when subclassing them (and probably more subtle issues). For example right now you can use ```super``` in a subclass of ```Validator```, but not in a subclass of ```BasicErrorHandler```.

### DIFF
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -86,7 +86,7 @@ SCHEMA_ERROR_MISSING = "validation schema missing"
 """ Error representations """
 
 
-class ValidationError:
+class ValidationError(object):
     """ A simple class to store and query basic error information. """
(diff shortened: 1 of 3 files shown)


funkyfuture:
ah, yeah. coming from Python 3 i missed this. is it also a good practice for mixin classes?

dkellner:
I think so, otherwise you will be mixing old-style and new-style classes. And as far as i know in Python 3 all classes behave like new-style classes which is another good reason for doing this, so we don't have different behaviour in Python 2.

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/cerberus/commit/bacce0af69fda777afd34df339f270cbd613941a

Thanks!

PR #273, opened 2016-10-03
Tests updates

### LINKED ISSUE #213: Proposal: completely switch to py.test

funkyfuture:
this should remove the remainings of unittests and thus resolve #213.

yeah, i know, there is no joy in reviewing this. still, it's essential.

one thing that still can be done is to extend all calls to `assert_fail` with the `error` or `errors` argument. even better, make one of these mandatory.

another thing may be to condense all tests against the `sample_schema` with simple test questions by parametrization.


nicolaiarocci:
Excellent.
Merged after rebase at https://github.com/nicolaiarocci/cerberus/commit/2f22734c95a3b702a99231e57600ed16d4a4f7e0

PR #275, opened 2016-10-18
Added a paragraph to the Usage documentation

gramotei:
It is about that normalization and coercision is done on a copy of the
document and not the original. Right now it is not obvious


nicolaiarocci:
Thanks!

PR #282, opened 2016-11-12
`readonly` conflicts with `default` ("simple" fix)

dkellner:
This is a careful fix for issue #268 without introducing any new features (like in #272) and without any (known) breaking changes.

For the full discussion see #272.

### DIFF
--- a/cerberus/validator.py
+++ b/cerberus/validator.py
@@ -378,6 +378,15 @@ class Validator(object):
         return self._config.get('is_child', False)
 
     @property
+    def is_normalized(self):
+        """ ``True`` if the document is already normalized. """
+        return self._config.get('is_normalized', False)
+
+    @is_normalized.setter
+    def is_normalized(self, value):
+        self._config['is_normalized'] = value
+
+    @property
     def purge_unknown(self):
         """ If ``True`` unknown fields will be deleted from the document
             unless a validation is called with disabled normalization.
@@ -495,6 +504,10 @@ class Validator(object):
         self.__normalize_rename_fields(mapping, schema)
         if self.purge_unknown:
             self._normalize_purge_unknown(mapping, schema)
+        # Check `readonly` fields before applying default values because
+        # a field's schema definition might contain both `readonly` and
+        # `default`.
+        self.__validate_readonly_fields(mapping, schema)
         self.__normalize_default_fields(mapping, schema)
         self._normalize_coerce(mapping, schema)
         self.__normalize_containers(mapping, schema)
@@ -665,6 +678,12 @@ class Validator(object):
             mapping[new_name] = mapping[field]
             del mapping[field]
 
+    def __validate_readonly_fields(self, mapping, schema):
+        for field in (x for x in schema if x in mapping and
+                      self._resolve_rules_set(schema[x]).get('readonly')):
+            self._validate_readonly(schema[field]['readonly'], field,
+                                    mapping[field])
+
     def __normalize_default_fields(self, mapping, schema):
         fields = [x for x in schema if x not in mapping or
                   mapping[x] is None and not schema[x].get('nullable', False)]
@@ -738,6 +757,7 @@ class Validator(object):
         self.__init_processing(document, schema)
         if normalize:
             self.__normalize_mapping(self.document, self.schema)
+            self.is_normalized = True
 
         for field in self.document:
             if self.ignore_none_values and self.document[field] is None:
@@ -1038,8 +1058,17 @@ class Validator(object):
     def _validate_readonly(self, readonly, field, value):
         """ {'type': 'boolean'} """
         if readonly:
-            self._error(field, errors.READONLY_FIELD)
-            return True
+            if not self.is_normalized:
+                self._error(field, errors.READONLY_FIELD)
+                return True
+            # If the document was normalized (and therefore already been
+            # checked for readonly fields), we still have to return True
+            # if an error was filed.
+            has_error = errors.READONLY_FIELD in \
+                self.document_error_tree.fetch_errors_from(
+                    self.document_path + (field,))
+            if self.is_normalized and has_error:
+                return True
 
     def _validate_regex(self, pattern, field, value):
         """ {'type': 'string'} """
--- a/docs/normalization-rules.rst
+++ b/docs/normalization-rules.rst
@@ -57,6 +57,8 @@ subdocuments like ``allow_unknown`` (see :ref:`allowing-the-unknown`). The defau
 
 .. versionadded:: 1.0
 
+.. _default-values:
+
 Default Values
 --------------
 You can set default values for missing fields in the document by using the ``default`` rule.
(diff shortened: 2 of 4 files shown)


cerberus/validator.py
```diff
@@ -665,6 +678,12 @@ def _normalize_rename_handler(self, mapping, schema, field):
             mapping[new_name] = mapping[field]
             del mapping[field]
 
+    def __validate_readonly_fields(self, mapping, schema):
+        for field in [x for x in schema if x in mapping and
```
funkyfuture: i'd prefer a generator- over a list-comprehension here.

cerberus/validator.py
```diff
@@ -1038,8 +1058,17 @@ def _validate_keyschema(self, schema, field, value):
...
         if readonly:
-            self._error(field, errors.READONLY_FIELD)
-            return True
+            if not self.is_normalized:
+                self._error(field, errors.READONLY_FIELD)
+                return True
+            # If the document was normalized (and therefore already been
+            # checked for readonly fields), we still have to return True
+            # if an error was filed if readonly is a priority rule.
+            if 'readonly' in self.priority_validations:
```
funkyfuture: isn't that always true? or is this different in the eve-validator?

cerberus/validator.py
```diff
@@ -1038,8 +1058,17 @@ def _validate_keyschema(self, schema, field, value):
...
+                has_error = self.schema_error_tree.fetch_errors_from(
```
funkyfuture: shall any (or a prior readonly) error be respected here?

cerberus/validator.py
dkellner: No, I did not think about Eve at all when implementing this. There is this "configuration" for priority validations, so I thought it's best to reuse it instead of hardcoding the fact. But thinking about that again it's special anyway because returning True is just something priority validations do. I will remove that check.

cerberus/validator.py
dkellner: Just a prior readonly error on this particular field. I think it's more clean and specific when the readonly-logic just cares about it's own errors. Do you see any problems with that approach?

cerberus/validator.py
dkellner: You're right, I will change that.

cerberus/validator.py
funkyfuture: no problem with the approach, but that would be `errors.READONLY_FIELD in self.schema_error_tree...`.

dkellner: Are you sure?  I thought the `SchemaErrorTree` allows to access errors for certain rules. In fact, I just tried that and indeed only get the error for `readonly`:

```python

d = errors.DocumentErrorTree()
s = errors.SchemaErrorTree()

e = errors.ValidationError(('a'), ('a', errors.READONLY_FIELD.rule),
    errors.READONLY_FIELD.code, errors.READONLY_FIELD.rule,
    None, None, None)
f = errors.ValidationError(('a'), ('a', errors.MIN_VALUE.rule),
    errors.MIN_VALUE.code, errors.MIN_VALUE.rule,
    None, None, None)
[tree.add(error) for tree in (d, s) for error in (e, f)]

# DocumentErrorTree returns all errors for the given field.
print(d.fetch_errors_from(('a',)))
>> [ValidationError @ 0x7f7686359ed0 ( document_path=a,schema_path=('a', 'min'),code=0x42,constraint=None,value=None,info=None ), ValidationError @ 0x7f76874f2150 ( document_path=a,schema_path=('a', 'readonly'),code=0x63,constraint=None,value=None,info=None )]

# SchemaErrorTree returns just the readonly error.
print(s.fetch_errors_from(('a', 'readonly'))))
>> [ValidationError @ 0x7f76874f2150 ( document_path=a,schema_path=('a', 'readonly'),code=0x63,constraint=None,value=None,info=None )]
```

funkyfuture:
some elaboration on this in the docs would be good.

cerberus/validator.py
funkyfuture: shower thought: wouldn't a query on the document error tree be safer with regards to implicit validation like from `allow_unknown`?

nicolaiarocci:
Very nice. Agree that a docs upgrade would be good.

cerberus/validator.py
dkellner: I'm not sure if an unknown field can ever be `readonly` in a meaningful way. I find querying the schema tree much clearer to understand, but that's maybe just my way of thinking. If you agree I would like to keep it this way :).

dkellner:
Yep, I will update the docs later today and update this PR so we can hopefully close this issue soon.

cerberus/validator.py
funkyfuture: i have no preference here. you are right that the example case is hardly meaningful, but people come up with uses the issue tracker i'd never thought of regularly.
my way of thinking here is: schemas can be tricky constructs and sometimes rules are evaluated that have not be adressed directly. documents on the other hand are clearer, in the document error tree you will find all errros associated with a field at exactly one node and nowhere else. it seems less error prone to me for the rather weird cases. and as mentioned above, someone will come up with something weird (which is charming).
why would you say the scheme error tree is clearer to understand? genetically they are the same.

dkellner: I did not mean my understanding of both trees, but the code readability at that specific location. But we don't need a deep discussion about this small issue, I am happy to go with your suggestion as well - you are way more experienced with the Cerberus codebase than I am, so I trust you with that.

dkellner:
Looks like the pypy3 test fails for some technical reason. @nicolaiarocci: Can you fix that?

docs/normalization-rules.rst
```diff
@@ -90,6 +90,14 @@ string, it points to a :doc:`custom method <customize>`.
    >>> v.errors
    {'a': ["default value for 'a' cannot be set: Circular dependencies of default setters."]}
 
+You can even use both `default` and `readonly` on the same field. This will
```
funkyfuture: don't let this bit delay this pr to be merged, but a link do the other doc on each side would be handy.
other rule names are commonly enclosed by double backticks for monospaced appearance.
is the feature/fix worthenough to extend the docs?

docs/validation-rules.rst
```diff
@@ -427,6 +427,9 @@ present in the target dictionary. This is useful, for example, when receiving
 a payload which is to be validated before it is sent to the datastore. The field
 might be provided by the datastore, but should not writable.
 
+.. versionchanged:: 1.0.2
+   Can be used in conjunction with `default`.
```
funkyfuture: and `default_setter`

funkyfuture:
[this issue] may be of interested concerning PyPy3 on travis.

the test succeed locally.

docs/normalization-rules.rst
dkellner: I will look into that. First I finally have to setup Sphinx to compile the docs locally ;).

nicolaiarocci:
I temporarily removed pypy3 from CI matrix, I'll just test PRs locally until it's fixed upstream.

nicolaiarocci:
Merged with rebase: https://github.com/nicolaiarocci/cerberus/commit/7408da5be6286ad9a849d859c8fbc31cf616529b

Thanks!

Issue #274, opened 2016-10-14
Problem with required and dependencies


evga:
Used Cerberus version 1.0.1 installed with pip
- I consulted these documentations:
  - [ Validation Rules] http://docs.python-cerberus.org/en/stable/validation-rules.html
#### Support request / Bug report

According to the docs, the following program should validate successfully since 
the `auth-key` field is required only when the `auth` field is `wep` or `wpa`.

``` python

import cerberus

schema = {
    'auth' : {
        'type':'string',
        'required':True,
        'allowed':['open','wep','wpa']
    },
    'auth-key' : {
        'type':'string',
        'required':True,
        'dependencies': {'auth': ['wep', 'wpa']}
    }
}

v = cerberus.Validator(schema)
print v.validate({'auth':'open'})
print v.errors
```

The output of the program is the following:

```
False
{'auth-key': ['required field']}
```

Is this a bug ?

gtnx:
I don't think it's a bug. You need to be more precise when specifying the auth-key:

```python
schema = {
    'auth' : {
        'type':'string',
        'required':True,
        'allowed':['open','wep','wpa']
    },
    'auth-key' : {
        'type':'string',
        'oneof': [{
            'required':True,
        }, {
            'dependencies': {'auth': ['open']},
        }]
    }
}
```

It's like saying: "wether auth-key is defined or auth is 'open'".

aguirrel:
@gtnx I think this is a bug or documentation is not ok, I have the same situation.

drcraig:
I think I'm also hitting this issue, too. The [documentation](http://docs.python-cerberus.org/en/stable/validation-rules.html#required) says 

> Also, if dependencies are declared for the field, its required rule will only be validated if all dependencies are included with the document.

In @aguirrel's example, I interpret that to mean that auth-key will only be required if auth's value is 'open'. Is that the intended behavior or am I misunderstanding?

aguirrel:
@drcraig, in @evga example, auth-key field should be required if value is 'wep' or 'wpa' (dependencies).

aguirrel:
I'm doing a regression test and it seems that in version 0.9 @evga example works fine. It's broken from 0.9.1 and above.

drcraig:
@aguirrel Right, my mistake. I was projecting my own situation where my dependency is only one particular value. Thanks for running the regression tests.

funkyfuture:
[...]

Issue #229, opened 2016-04-25
extend documentation with advanced usages


funkyfuture:
there are repeated requests for support with special use cases. mostly these are solvable with one of the `*of`-rules. currently open issues are #221, #194, #182 and #136.

it seems advisable to add a page to the docs that covers more complex uses of cerberus by examples. beside the fact that i'm not really keen on solving other people's detailed problems, it also seems advisable that these examples are written from a user's perspective. but certainly i'm willing to support this by giving hints and reviewing.

has anyone an idea how to encourage people to share their experiences and solutions?

nicolaiarocci:
Would also be a great occasion for people to contribute to an open source project. In Eve we have a [Snippets](http://python-eve.org/snippets/) section which is supposed to receive user snippets and use cases. Maybe something like this would work? Can't say it got a lot of traction in Eve.

funkyfuture:
it's not the case that i don't have strong arguments against the following proposal, but i tend to close this issue and outsource any support questions to StackOverflow.

PR #286, opened 2016-11-22
Updating test cases to consider issue #274 - Problem with required an…

aguirrel:


cerberus/tests/test_validation.py
```diff
@@ -840,6 +840,7 @@ def test_dependencies_errors():
                 error=('field2', ('field2', 'dependencies'),
                        errors.DEPENDENCIES_FIELD_VALUE,
                        {'field1': ['one', 'two']}, ({'field1': 'three'},)))
+    assert_success({'field1': 'three'}, validator=v)
```
funkyfuture: this should not succeed as `field2` is defined to be required.

aguirrel: I have just modified source code to be compliant with docs.  All test cases are working now, included this one ;)

funkyfuture:
my impression was rather that the docs should be clarified. but that should be discussed in #229.

i'd had some minor remarks here. and, obviously, the docs would have to be amended if this was changed in a very elaborative and exemplary way. :-)

aguirrel:
ok I understand. Please, correct me if I'm wrong, but I think people base theirs development on docs (at least two persons did ;) ).  Indeed I think is a very consistent way to manage 'required' validator.  In fact this new test case worked on version 0.9 but not in 0.9.1.

I think merging this PR is like "fixing" an issue but if you rewrite docs it could be seem a little odd.

Personally I suggest not changing so much in docs in version 1.0.x and change it in 1.1.x.

funkyfuture:
afaik, the behavior you expect was never intended, and as pointed out in the correspoinding issue, the docs can be interpreted like you do, but are not meant to. for an informed discussion it would be good to know which commit changed that behaviour. `git bisect` is great for that.

aguirrel:
Ok I understand.  Please feel free to to close this PR without merging.  Anyway I can't do it work in all situations.  If in the test case, you just invert the order of the tests (first success and then fail) it just doesn't work.  It seems that this PR only works if the field is present in the document you are validating or you name it in other validator call (as is happening in the test case).

Issue #280, opened 2016-11-11
Handling unallowed unicode value


inirudebwoy:
Used Cerberus version: 1.0.1

- I consulted these documentations:
  - [x] http://docs.python-cerberus.org/en/stable/
  - [x] http://docs.python-cerberus.org/en/latest/

- [ ] I found nothing relevant to my problem in the docs.
- [ ] I found the documentation not helpful to my problem.

- [x] I have the capacity to improve the docs when my problem is solved.
- [x] I have the capacity to submit a patch when a bug is identified.

I'm using Cerberus to validate data send by users, data can be unicode. One of the validated fields needs to be a certain `allowed` value. Value itself can be a unicode string but `allowed` list does not have any unicode elements.

This is a simple illustration of the problem.
```python
from cerberus import Validator
schema = {'name': {'type': 'string', 'allowed': ['Joe']}
v = Validator(schema)

In [1]: v.validate({'name': 'Jo'})
Out[1]: False

In [2]: v.errors
Out[2]: {'name': ['unallowed value Jo']}

In [3]: v.validate({'name': u'Michał'})
Out[3]: False

In [4]: v.errors
---------------------------------------------------------------------------
/home/majki/.virtualenvs/cerberus-playground/lib/python2.7/site-packages/cerberus/errors.pyc in format_message(self, field, error)
    462         return self.messages[error.code]\
    463             .format(*error.info, constraint=error.constraint,
--> 464                     field=field, value=error.value)
    465
    466     def insert_error(self, path, node):

UnicodeEncodeError: 'ascii' codec can't encode character u'\u0142' in position 5: ordinal not in range(128)

```
As you can see an exception occurs when I try to display the error message. 
At first I was encoding the data before validation, what worked, but than it would require to decode it back before rendering a response. So it seemed a bit off to do.

Since my `allowed` list does not contain any unicode strings as a workaround I have created a validator that rejects non ASCII string with a proper message and does manual check for `allowed` values.

Is this a bug? Or am I doing something horribly wrong with unicode?

funkyfuture:
though this could propably be fixed, i suggest you use Python 3.

nicolaiarocci:
What @funkyfuture said. However it would be nice if Cerberus would run smoothly over this, even on Py2. So @inirudebwoy, feel free to submit a patch 😄

inirudebwoy:
Cheers guys.
Since I can't unfortunately move to Python 3, will have to submit a patch.

PR #287, opened 2016-11-22
Handle unicode in 'allowed' 

inirudebwoy:
Hi,
This is a try to fix issue #280. Differences in how Py3 and Py2 are handling the `str` makes the `format_message` method bit too complicated.
Py3 is fine and can handle anything in terms of string handling which can't be said about Py2.
This difference unfortunately makes following test to fail on Py3.
```
doc = {'letters': u'♄εℓł☺'}
schema = {'letters': {'type': 'string', 'allowed': ['♄εℓł☺']}}

# tested value is Unicode
# on Python2 this is failure as 'allowed' is encoded with file coding
# on Python3 this is success as 'allowed' is a py3 str
assert_fail(doc, schema)
```

How should I handle this? Does it matter? 
This case possibly could be an invalid code or bad practice as str will be encoded by Python with the defined encoding making this part a bit ambiguous. I'd expect one to encode unicode endpoints explicitly or use the unicode object.

Also I decided to catch the exception that indicates code runs on Py3 instead of using six module, as you do not have any dependencies.

### DIFF
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -5,9 +5,10 @@ from __future__ import absolute_import
 
 from collections import defaultdict, namedtuple, MutableMapping
 from copy import copy, deepcopy
+from functools import wraps
 from pprint import pformat
 
-from cerberus.platform import py2_error_unicode_fix
+from cerberus.platform import PYTHON_VERSION
 from cerberus.utils import compare_paths_lt, quote_string
 
 
@@ -390,6 +391,33 @@ class ToyErrorHandler(BaseErrorHandler):
         pass
 
 
+def encode_unicode(f):
+    """Cerberus error messages expect regular binary strings.
+    If unicode is used in a ValidationError message can't be printed.
+
+    This decorator ensures that if legacy Python is used unicode
+    strings are encoded before passing to a function.
+    """
+    @wraps(f)
+    def wrapped(obj, error):
+
+        def _encode(value):
+            """Helper encoding unicode strings into binary utf-8"""
+            if isinstance(value, unicode):  # noqa: F821
+                return value.encode('utf-8')
+            return value
+
+        error = copy(error)
+        error.document_path = _encode(error.document_path)
+        error.schema_path = _encode(error.schema_path)
+        error.constraint = _encode(error.constraint)
+        error.value = _encode(error.value)
+        error.info = _encode(error.info)
+        return f(obj, error)
+
+    return wrapped if PYTHON_VERSION < 3 else f
+
+
 class BasicErrorHandler(BaseErrorHandler):
     """ Models cerberus' legacy. Returns a :class:`dict`. """
     messages = {0x00: "{0}",
@@ -448,7 +476,7 @@ class BasicErrorHandler(BaseErrorHandler):
     def __str__(self):
         return pformat(self.pretty_tree)
 
-    @py2_error_unicode_fix
+    @encode_unicode
     def add(self, error):
         if error.is_logic_error:
             self.insert_logic_error(error)
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -1,3 +1,4 @@
+# -*-: coding utf-8 -*-
 """ This module contains the error-related constants and classes. """
 
 from __future__ import absolute_import
@@ -459,9 +460,15 @@ class BasicErrorHandler(BaseErrorHandler):
         self.tree = {}
 
     def format_message(self, field, error):
-        return self.messages[error.code]\
-            .format(*error.info, constraint=error.constraint,
-                    field=field, value=error.value)
+        message = self.messages[error.code]
+        try:
+            if isinstance(error.value, unicode):
+                message = unicode(message)
+        except NameError:
+            pass
+
+        return message.format(*error.info, constraint=error.constraint,
+                              field=field, value=error.value)
 
     def insert_error(self, path, node):
         """ Adds an error or sub-tree to :attr:tree.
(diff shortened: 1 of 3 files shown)


cerberus/errors.py
```diff
@@ -459,9 +460,15 @@ def clear(self):
         self.tree = {}
 
     def format_message(self, field, error):
-        return self.messages[error.code]\
-            .format(*error.info, constraint=error.constraint,
-                    field=field, value=error.value)
+        message = self.messages[error.code]
+        try:
+            if isinstance(error.value, unicode):
```
funkyfuture: also, `constraint`, `field` and members of `info` could be unicode strings.

cerberus/errors.py
```diff
@@ -459,9 +460,15 @@ def clear(self):
...
+                message = unicode(message)
+        except NameError:
```
funkyfuture: this makes it rather look like a workaround.

inirudebwoy: Agree, since it's code specific to Python version implicit check for current Python version would be better. Unless this part targeting the version could be moved outside to `platform` module.

inirudebwoy:
I agree with porting remark and would gladly do it. It's not really possible at this point in time (possibly next year) and not really necessary as I have a workaround for this. The list of allowed values does not include any unicode strings thus I can reject anything that is unicode.
This is my situation, what I'm hoping to achieve is to fix this problem so other won't have problems. I understand that if no one has reported this problem since 2012 and probably today more people use Cerberus with Python3 (potentially because of Unicode) it may not be worth fixing and adding complexity.

`platform` module could be a good place to keep part of version specific differences. 

`add` method calls `insert_error` which has the output of `format_message` as one of the arguments.
I've put my changes into `format_message` as it's names describes best what my changes are doing. `insert_error` nor `add` deal with the message, they handle appending the message itself into proper place within the structure. Other place I was considering is the `ValidationError` class, possibly modifying the values on access.

funkyfuture:
> Other place I was considering is the ValidationError class, possibly modifying the values on access.

liked the idea, but it's a problem specific to the `BasicErrorHandler` as it represents error information as messages (strings). `ValidationError` itself shouldn't be altered as this would falsify error information. i think a decorator around `add` that wraps  it with a converter (that returns a *copy* of an error) in case of Python 2 would be good. mind that the messages-attribute alse needs conversion.

inirudebwoy:
> to convert in format_message would miss dictionary keys of the tree, or not?

Yes, indeed. `add` allows to replace the whole `ValidationError` with a copy.

Sorry for late reply, I have not forgotten about this PR. By the end of this week I'll have it cleaned up. 
I have a working prototype now with changes you recommended. It works better as it handles all the cases, even if user had used encoded strings as values or used them in schema definition.

nicolaiarocci:
I'm eager to release a new version. What's the state of this PR? Should I close it for the time being, or is work being done on it?

inirudebwoy:
Sorry. Yes I'm working on it. I'll update it in couple days. If you want to release sooner than that please close it.

cerberus/platform.py
```diff
@@ -12,3 +14,31 @@
 else:
     _str_type = str
     _int_types = (int,)
+
+
+def py2_error_unicode_fix(f):
+    """Cerberus error messages expect regular binary strings.
+    If unicode is used in schema message can't be printed.
```
funkyfuture: 'schema' -> a `ValidationError` attribute

cerberus/platform.py
```diff
@@ -12,3 +14,31 @@
 else:
     _str_type = str
     _int_types = (int,)
+
+
+def py2_error_unicode_fix(f):
+    """Cerberus error messages expect regular binary strings.
```
funkyfuture: ... on Python 2

cerberus/platform.py
```diff
@@ -12,3 +14,31 @@
...
+        if PYTHON_VERSION < 3:
+            error = copy.copy(error)
+            error.document_path = _encode(error.document_path)
+            error.schema_path = _encode(error.schema_path)
+            error.constraint = _encode(error.constraint)
+            error.value = _encode(error.value)
+            error.info = _encode(error.info)
+
+        return f(obj, error)
+    return wrapped
```
funkyfuture: return wrapped if PYTHON_VERSION < 3 else f

cerberus/platform.py
```diff
@@ -12,3 +14,31 @@
...
+def py2_error_unicode_fix(f):
+    """Cerberus error messages expect regular binary strings.
+    If unicode is used in schema message can't be printed.
+
+    This decorator ensures that if legacy Python is used unicode
+    strings are encoded before passing to a function.
+    """
+    @wraps(f)
+    def wrapped(obj, error):
+        if PYTHON_VERSION < 3:
```
funkyfuture: this should only be checked when the decorator is processed, see below.

cerberus/platform.py
```diff
@@ -12,3 +14,31 @@
 else:
     _str_type = str
     _int_types = (int,)
+
+
+def py2_error_unicode_fix(f):
```
funkyfuture: i think function names shouldn't have 'fix' in their names.

cerberus/platform.py
```diff
@@ -12,3 +14,31 @@
...
+            error.value = _encode(error.value)
+            error.info = _encode(error.info)
+
+        return f(obj, error)
+    return wrapped
+
+
+def _encode(value):
+    """Helper encoding unicode strings into binary utf-8"""
+    if isinstance(value, unicode):
```
funkyfuture: append `# noqa: F821` here

funkyfuture:
lgtm, thanks!

nicolaiarocci:
Merged after rebase, see https://github.com/nicolaiarocci/cerberus/commit/d9df40850c30fbde7b4ee8570d881d4a5e54a3d9

Thanks!

PR #289, opened 2016-11-23
Fix custom coercers documentation

### LINKED ISSUE #285: Undefined variable in `Custom Coercers` snippet from documentation
Used Cerberus version: 1.0.1

  - [x] http://docs.python-cerberus.org/en/stable/
  - [X] http://docs.python-cerberus.org/en/latest/

  - http://docs.python-cerberus.org/en/latest/customize.html#custom-coercers

- [X] I found the documentation not helpful to my problem.

- [X] I have the capacity to improve the docs when my problem is solved.
- [X] I have the capacity to submit a patch when a bug is identified.

----

#### Use-case abstract

I want to be able to set finer error messages when defining custom coercers.

----

Here is the excerpt from the documentation (http://docs.python-cerberus.org/en/latest/customize.html#custom-coercers):

```
class MyNormalizer(Validator):
        def __init__(self, multiplier, *args, **kwargs):
            super(MyNormalizer, self).__init__(*args, **kwargs)
            self.multiplier = multiplier

        def _normalize_coerce_multiply(self, value):
            try:
                return value * self.multiplier
            except Exception as e:
                self._error(field, errors.COERCION_FAILED, e)
```

This code snippet can't work since `field` variable in the last line is undefined. We either need to pass the `field` argument to the `_normalize_coerce_multiply` method or avoid the try/except. Correct me if I am wrong, but I reckon that the try/except is already done in the `__normalize_coerce` method which would mean that it's just a documentation issue.

Ideally, I would like to be able to define custom error messages in coercers, and generally be finer in the way that I handle my error messages using Cerberus. I have already encountered the same kind of limitation with 1.x custom types. In 0.9, there was no limitation in this regard (`field` argument was passed to the child method, but that was removed in 1.0 for some reason). This actually prevents me from migrating to Cerberus 1.x in one of my projects (I relied on `_error` method in a custom type). I think it would be a good idea to pass the `field` argument everywhere (types, coercers, validators, …) if requested by  the child method (maybe using argspec or something similar to prevent breaking changes). But that's maybe a bit out of the scope of this issue.

Many thanks for the amazing work :) .

Gilb's

gilbsgilbs:
Closes #285.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,3 +43,4 @@ Patches and Suggestions
 - Trong Hieu HA
 - Waldir Pimenta
 - calve
+- gilbsgilbs
--- a/docs/customize.rst
+++ b/docs/customize.rst
@@ -127,10 +127,7 @@ a method as ``rename_handler``. The method name must be prefixed with
             self.multiplier = multiplier
 
         def _normalize_coerce_multiply(self, value):
-            try:
-                return value * self.multiplier
-            except Exception as e:
-                self._error(field, errors.COERCION_FAILED, e)
+            return value * self.multiplier
 
 .. doctest::
 


funkyfuture:
lgtm, thanks.

nicolaiarocci:
Thanks!

PR #294, opened 2017-01-04
Fixes normalization with definitions from registries

funkyfuture:


funkyfuture:
interesting, `test_validated_schema_cache` produces a different amounts of `cache_size` for different interpreters. though no clear distinction between Python 2 and 3.

PR #296, opened 2017-01-18
.dockerignore actually supports **-globbing

dkellner:
see https://docs.docker.com/engine/reference/builder/#/dockerignore-file

Quote:
> Beyond Go’s filepath.Match rules, Docker also supports a special wildcard string ** that matches any number of directories (including zero). For example, **/*.go will exclude all files that end with .go that are found in all directories, including the root of the build context.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,22 +1,10 @@
 .*
-*.egg-info
-*.orig
-*.pyc
-__pycache__
+**/*.egg-info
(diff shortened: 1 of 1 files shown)


funkyfuture:
cool, finally.

lgtm, thanks.

nicolaiarocci:
thanks!

Issue #271, opened 2016-09-27
Get error when I set fields as tuples instead of lists


theosotr:
I have the following piece of code.

``` python
from cerberus import Validator


schema = {
    'my_field': {
        'type': 'list',
        'schema': {'type': 'string'}
    }
}

validator = Validator(schema)

data = {'my_field': ('foo', 'bar')}

validator.validate(data)
```

I get the following exception:
`TypeError: 'tuple' object does not support item assignment`

When I set `my_field` as list works fine. However, I would like cerberus to support `tuple objects` as well.

funkyfuture:
could you please provide the full traceback?!

andreymal:
I'm not author, but i could:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/andreymal/.local/lib/python3.5/site-packages/cerberus/validator.py", line 739, in validate
    self.__normalize_mapping(self.document, self.schema)
  File "/home/andreymal/.local/lib/python3.5/site-packages/cerberus/validator.py", line 499, in __normalize_mapping
    self.__normalize_containers(mapping, schema)
  File "/home/andreymal/.local/lib/python3.5/site-packages/cerberus/validator.py", line 564, in __normalize_containers
    self.__normalize_sequence(field, mapping, schema)
  File "/home/andreymal/.local/lib/python3.5/site-packages/cerberus/validator.py", line 620, in __normalize_sequence
    mapping[field][i] = result[i]
TypeError: 'tuple' object does not support item assignment
```

-------

Generally, Cerberus modifies input, i think this is very bad:

```python
original = [1, 2]

schema = {
    'my_field': {'type': 'list', 'schema': {'type': 'string', 'coerce': str}}
}

validator = Validator(schema)
validator.validate({'my_field': original})

print(original == [1, 2])  # False — WAT?
print(original == ['1', '2'])  # True — WAAT??!1
```

funkyfuture:
@theosotr support for normalizing tuples would have to be amended. it's certainly welcome. in your case you could also avoid hitting the normalization code by calling `validate` with `normalize=False`.

@andreymal this behavior has recently been amended to the docs.

PR #301, opened 2017-03-17
Normalising tuples support #271

elecay:


cerberus/validator.py
```diff
@@ -666,8 +666,11 @@ def __normalize_sequence(self, field, mapping, schema):
             document_crumb=field, schema_crumb=(field, 'schema'),
             schema=schema)
         result = validator.normalized(document, always_return_document=True)
-        for i in result:
-            mapping[field][i] = result[i]
+        if len(result.values()) > 0:
```
funkyfuture: is that test necessary? `if results:` would be more idiomatic.

cerberus/validator.py
```diff
@@ -666,8 +666,11 @@ def __normalize_sequence(self, field, mapping, schema):
...
+            if type(mapping[field]) is tuple:
```
funkyfuture: that wouldn't work with subclasses of `tuple`, therefore the `isinstance` function is common to identify types.
nonetheless, it should be enough to create a new object of a particular type: `mapping_field = type(mapping[field])(result.values)`

funkyfuture:
oh, and for fixing pull requests i recommend to [squash your commits](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history#squash-commits-together) (or rather fix) and then force-push your branch.

cerberus/validator.py
elecay: mind-blowing for me

elecay:
Awesome comments. Will fix this.

cerberus/validator.py
funkyfuture: don't get addicted or you'll find yourself using Python for the laundry. ;-)

cerberus/validator.py
```diff
@@ -666,8 +666,8 @@ def __normalize_sequence(self, field, mapping, schema):
             document_crumb=field, schema_crumb=(field, 'schema'),
             schema=schema)
         result = validator.normalized(document, always_return_document=True)
-        for i in result:
-            mapping[field][i] = result[i]
+        if result:
```
funkyfuture: with my previous comment i meant two things, sorry for being unclear:
- this expression is semantically identical to the previous version
- the expression shouldn't be needed at all, bit i might be wrong with that assumption

elecay: You are totally right. Sorry about that, and thank you for your guide and your time. 😄

funkyfuture:
i tested it against a  more nasty sequence and still found it to work. seems that value views are sorted in all Python version, which i didn't expect.

we could also use that type instantiation for the normalized mappings to maintain the input's type.

elecay:
Great. Should I do something else or can we consider this issue closed?

funkyfuture:
i'd consider it solved. i'll make a follow-up for the mappings later.

nicolaiarocci:
Thanks!

nicolaiarocci:
Closes #271

Issue #302, opened 2017-03-23
Regression: Allow unknown causes dictionary validation to fail with a ``KeyError``


daviskirk:
Used Cerberus version / latest commit: 1.1

- I consulted these documentations:
  - [x] http://docs.python-cerberus.org/en/stable/
  - [x] http://docs.python-cerberus.org/en/latest/  

- I consulted these sections of the docs (add more lines as necessary):
  - http://docs.python-cerberus.org/en/latest/changelog.html?highlight=changelog

- [ ] I found nothing relevant to my problem in the docs.
- [ ] I found the documentation not helpful to my problem.

- [x] I have the capacity to improve the docs when my problem is solved.
- [x] I have the capacity to submit a patch when a bug is identified.

----

#### Use-case abstract

Validating a dictionary with ``allow_unknown`` set to ``True``

----

#### Support request / Bug report

This might be a regression, since testing this on 1.0.1 works fine:

```python
v = cerberus.Validator({'a': {'type': 'dict', 'allow_unknown': True}})
v.validate({'a': {}})
```
Errors out with the exception:
```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "./lib/python3.5/site-packages/cerberus/validator.py", line 795, in validate
    self.__normalize_mapping(self.document, self.schema)
  File "./lib/python3.5/site-packages/cerberus/validator.py", line 549, in __normalize_mapping
    self.__normalize_containers(mapping, schema)
  File "./lib/python3.5/site-packages/cerberus/validator.py", line 607, in __normalize_containers
    field, mapping, schema)
  File "./lib/python3.5/site-packages/cerberus/validator.py", line 653, in __normalize_mapping_per_schema
    schema=schema[field]['schema'],
KeyError: 'schema'
```

The example without the ``'allow_unknown'`` key or with an invalid type work without error:
``` python
v = cerberus.Validator({'a': {'type': 'dict'}})
v.validate({'a': {}}) # True
v = cerberus.Validator({'a': {'type': 'dict', 'allow_unknown': True}})
v.validate({})  # True
v.validate({'a': 4})  # False
```

nicolaiarocci:
Thanks. Should be fixed now.

PR #303, opened 2017-03-24
Fix for issue 302

funkyfuture:
a schematic that solely relates to this pr's id:


nicolaiarocci:
Fix for #302

nicolaiarocci:
That's the Roland TB303 mainboard right?

funkyfuture:
Yep, I'm always excited when I somehow get across the number due to it.

PR #304, opened 2017-03-28
Override UnvalidatedSchema.copy

peterdemin:
because it doesn't have .validator attribute

### DIFF
--- a/cerberus/schema.py
+++ b/cerberus/schema.py
@@ -212,6 +212,11 @@ class UnvalidatedSchema(DefinitionSchema):
     def validate(self, schema):
         pass
 
+    def copy(self):
+        # Override ancestor's copy, because
+        # UnvalidatedSchema does not have .validator:
+        return self.__class__(self.schema.copy())
+
 
 class SchemaValidationSchema(UnvalidatedSchema):
     def __init__(self, validator):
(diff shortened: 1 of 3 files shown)


funkyfuture:
i'd recommend to add a regression test that implements a very basic version of your use-case.

out of interest, how much is the speed gain? about how many schemas are involved?

peterdemin:
How do I add a regression test? Should it be the same as my Unit test?
I don't remember the numbers, but we validate thousands of documents in a batch using pretty complex schema.

funkyfuture:
it's just a more practical test; rather a field experiment than one in the laboratory, one might coin it. the `test_customization.py` is full with such. the idea is, that your use-case might get broken by other changes that no one can predict atm.

wow, your scenario seems to be a success story. an article about it would be neat to get referenced in the docs.

peterdemin:
It's ok if a future version will not have the same way of overriding internals - I have my own set of tests to check compatibility. And I don't want to constraint cerberus refactorings.
While I like the idea of writing up our success story, it's going to take some time for me to get to it.
In the meantime, my PR is an obvious bug fix, and without it, I'm stuck on 1.0.1.

peterdemin:
@funkyfuture Did I get your idea right with customization test?

funkyfuture:
yep, and though i could follow your reasoning, it's still a good idea since i'll probably set things from head onto feet and make the validated schema a subclass of the unvalidated one.

nicolaiarocci:
Merged after rebase, see https://github.com/pyeve/cerberus/commit/82d79d1407b00a7373475dfe80f6340accedbd61

Thanks!

PR #306, opened 2017-04-13
Fix dependency is not considered satisfied if it has a null value.

### LINKED ISSUE #305: A dependency is not considered satisfied if it has a null value
Used Cerberus version / latest commit: 1.1

  - [X] http://docs.python-cerberus.org/en/stable/
  - [ ] http://docs.python-cerberus.org/en/latest/

  - http://docs.python-cerberus.org/en/stable/validation-rules.html#dependencies

- [ ] I found the documentation not helpful to my problem.

- [ ] I have the capacity to improve the docs when my problem is solved.

----

#### Use-case abstract

A dependency is not considered satisfied if it has a null value.

----

To reproduce:

```bash
>>> from cerberus import Validator
... validator = Validator(schema={
...     'field1': {
...         'required': False,
...         'nullable': True
...     },
...     'field2': {
...         'required': False,
...         'dependencies': 'field1'
...     }
... })
... document = {
...     'field1': None,
...     'field2': 'foo'
... }
... assert validator.validate(document), validator.errors
  File "<input>", line 16, in <module>
AssertionError: {'field2': ["field 'field1' is required"]}
```

This seems wrong considering required fields with a null value satisfy the required constraint, but not a dependency.

davidt99:
Closes #305 
It's also solves a case when a field is nullable and `none`, the dependencies requirements is ignored.
I'm not very happy with the solution but I don't that familiar with the code base.

### DIFF
--- a/cerberus/tests/test_validation.py
+++ b/cerberus/tests/test_validation.py
@@ -808,6 +808,30 @@ def test_dependencies_dict_with_required_field():
     assert_success({'foo': 'bar', 'bar': 'foo'}, schema)
 
 
+def test_dependencies_field_satisfy_nullable_field():
+    # https://github.com/pyeve/cerberus/issues/305
+    schema = {
+        'foo': {'required': False, 'nullable': True},
+        'bar': {'required': False, 'dependencies': 'foo'}
+    }
+
+    assert_success({'foo': None, 'bar': 1}, schema)
+    assert_success({'foo': None}, schema)
+    assert_fail({'bar': 1}, schema)
+
+
+def test_dependencies_field_with_mutually_dependent_nullable_fields():
+    schema = {
+        'foo': {'required': False, 'dependencies': 'bar', 'nullable': True},
+        'bar': {'required': False, 'dependencies': 'foo', 'nullable': True}
+    }
+    assert_success({'foo': None, 'bar': None}, schema)
+    assert_success({'foo': 1, 'bar': 1}, schema)
+    assert_success({'foo': None, 'bar': 1}, schema)
+    assert_fail({'foo': None}, schema)
+    assert_fail({'foo': 1}, schema)
+
+
 def test_dependencies_dict_with_subodcuments_fields():
     schema = {
         'test_field': {'dependencies': {'a_dict.foo': ['foo', 'bar'],
--- a/cerberus/validator.py
+++ b/cerberus/validator.py
@@ -93,7 +93,7 @@ class Validator(object):
     """ Rules that are evaluated on any field, regardless whether defined in
         the schema or not.
         Type: :class:`tuple` """
-    priority_validations = ('nullable', 'readonly', 'type')
+    priority_validations = ('dependencies', 'nullable', 'readonly', 'type')
     """ Rules that will be processed in that order before any other and abort
         validation of a document's field if return ``True``.
         Type: :class:`tuple` """
@@ -347,9 +347,9 @@ class Validator(object):
 
         parts = path.split('.')
         for part in parts:
-            context = context.get(part)
-            if context is None:
+            if part not in context:
                 return None, None
+            context = context[part]
 
         return parts[-1], context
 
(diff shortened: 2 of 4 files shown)


cerberus/tests/test_validation.py
```diff
@@ -808,6 +808,18 @@ def test_dependencies_dict_with_required_field():
     assert_success({'foo': 'bar', 'bar': 'foo'}, schema)
 
 
+def test_dependencies_dict_with_nullable_fields():
```
funkyfuture: please test the `nullable` issue in an extra test and add a reference to #305 as URL here.

cerberus/validator.py
```diff
@@ -97,6 +97,10 @@ class and a dictionary, the latter is passed to the
     """ Rules that will be processed in that order before any other and abort
         validation of a document's field if return ``True``.
         Type: :class:`tuple` """
+    post_priority_validation = ('dependencies',)
```
funkyfuture: though i introduced them, i'm not particular fond of these attributes, so i'd prefer not to introduce another one. i see that it has a small footprint, nonetheless i'd like to ask whether another approach is thinkable.

cerberus/validator.py
```diff
@@ -97,6 +97,10 @@ class and a dictionary, the latter is passed to the
...
+    """ Rules that will be processed if defined when one of the priority
```
funkyfuture: i do not exactly understand that sentence. in particular the 'if' followed by 'when' without a conjunction of the condition expressions (if they are that, i'm assuming).

`flake8` tells us, there's a trailing whitespace here.

cerberus/validator.py
```diff
@@ -862,6 +866,12 @@ def validate_rule(rule):
                              x in self.mandatory_validations))
         for rule in prior_rules:
             if validate_rule(rule):
```
funkyfuture: `if validate(rule) and not self._errors:`

davidt99:
@funkyfuture Changed the behavior, dependencies check are now part of the priority checks, but never aborts the validation.

cerberus/validator.py
```diff
@@ -899,10 +899,6 @@ def _validate_dependencies(self, dependencies, field, value):
         elif isinstance(dependencies, Mapping):
             self.__validate_dependencies_mapping(dependencies, field)
 
-        if self.document_error_tree.fetch_node_from(
```
davidt99: I didn't fully understand what this code did, but nobody was looking at the result until now, so I removed it in order to avoid aborting the validation check.

funkyfuture:
i submitted a pr that solely deals with #305 as it wasn't clear to what change fixed what in your pr and i needed clarity when reviewing.

concerning the second issue tackled here, we need to discuss a little more and find a proper solution:

- making `dependencies` a priority rule undermines Cerberus' paradigm to be non-blocking
- i find this hardly acceptable as this addresses some special cases, but affects any others as well
- funnily, the part you removed would be necessary to trigger an abortion

the problem here doesn't stem from the way how `dependencies` is evaluated, but that `nullable` is aborting a field's validation when `True` and the value is `None`.

since it's mainly issues coming up that have to do with the interaction of rules recently, i strongly object to address these by dealing with edge-cases in the seemingly simplest way in a rather hot-fixing manner. it makes the code harder to understand and to extend.
we have an architectural problem here as the validation procedure is quiet static while the interaction of rules requires something more dynamic.

davidt99:
It seems that I wasn't clear enough, so I'll try explain in more details:
I have an optional start and end parameter which if one appears, the other must too. `None` value represent infinity or minus infinity depends on which field it appears.
The problem is that the current code doesn't continue to validate dependencies after it finds the field `None`.
I added the dependencies to the priority to make sure dependencies will always be checked (which make sense because it doesn't matter if the field is `None`, readonly or wrongly typed).
The part that I deleted as far as I can tell didn't have effect on the validation, and I wanted to make sure dependencies won't abort validation in order to stay consistent with the current behavior.
I agree that there is an architectural problem, but I don't feel comfortable to make such big design changes.

funkyfuture:
just an idea, that may be investigated: we expose the pipelines `prior_rules` and `rules` in `Validator.__validate_definitions` in order to be manipualted if needed.

davidt99:
I think that before we proceed, we need to decided whether dependencies should be validated regardless of the other validation.
IMHO it should because dependencies are irrelevant to whether the field itself is valid or not.
Regarding your first suggestion, I think any solution that involve configuring the validator object and not the schema dictionary is not a good solution since it makes code reuse harder (my use case is to have a certain dictionary bound to a specific REST api and the same code checks every request).
Regarding your second suggestion (if I understood it correctly), there is more rules that need to be dropped: max, min, allowed, empty, regex and more.

PR #307, opened 2017-04-17
Fixes the lookup for dependency fields.

### LINKED ISSUE #305: A dependency is not considered satisfied if it has a null value

funkyfuture:
since it wasn't clear to me what parts of #306 are necessary to solve #305, i took a look at it and found that the relationship of `nullable` and `ignore_none_values` required some clarification.
the test schema is intentionally reduced from the ones @derek-miller and @davidt99 provided.

Fixes #305 and partly replaces #306.

### DIFF
--- a/cerberus/tests/test_validation.py
+++ b/cerberus/tests/test_validation.py
@@ -808,7 +808,19 @@ def test_dependencies_dict_with_required_field():
     assert_success({'foo': 'bar', 'bar': 'foo'}, schema)
 
 
-def test_dependencies_dict_with_subodcuments_fields():
+def test_dependencies_field_satisfy_nullable_field():
+    # https://github.com/pyeve/cerberus/issues/305
+    schema = {
+        'foo': {'nullable': True},
+        'bar': {'dependencies': 'foo'}
+    }
+
+    assert_success({'foo': None, 'bar': 1}, schema)
+    assert_success({'foo': None}, schema)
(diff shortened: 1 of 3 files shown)


davidt99:
LGTM. We can continue to discuss the other issue I encountered on #306 PR conversation page.

nicolaiarocci:
Merged after rebase. Thanks!

PR #309, opened 2017-04-18
Fixes depencies evaluation if nullable

funkyfuture:
(here it is with a clean git history, why must the git fiddle be played upon bedtime?)

spin-off of #306. cc @davidt99

@nicolaiarocci i wouldn't release this too soon, issues with custom validator may be ahead of us. cc @peterdemin

anyone reviewing this, are there further appliances?

### DIFF
--- a/cerberus/validator.py
+++ b/cerberus/validator.py
@@ -98,8 +98,8 @@ class Validator(object):
         validation of a document's field if return ``True``.
         Type: :class:`tuple` """
     _valid_schemas = set()
-    """ A :class:`set` of hashed validation schemas that are legit for a
-        particular ``Validator`` class. """
+    """ A :class:`set` of hashed derived from validation schemas that are 
+        legit for a particular ``Validator`` class. """
 
     def __init__(self, *args, **kwargs):
         """ The arguments will be treated as with this signature:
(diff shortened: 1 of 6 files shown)


nicolaiarocci:
Should we release 1.2 then, and leave this out for a later release?

funkyfuture:
nah, i'd like to see it in since that would bundle some improvements concerning `dependencies`, wouldn't it? what about releasing a release candidate in one week and a call for testing? and a final release two weeks after a rc was released and received no bug reports.

i really don't care about the times. can we figure out something reasonable that fits @peterdemin's expectations?

cerberus/validator.py
```diff
@@ -1065,11 +1085,12 @@ def _validate_minlength(self, min_length, field, value):
     def _validate_nullable(self, nullable, field, value):
         """ {'type': 'boolean'} """
         if value is None:
-            if nullable:
-                return True
-            else:
+            if not nullable:
                 self._error(field, errors.NOT_NULLABLE)
-                return True
+            self._drop_remaining_rules(
```
davidt99: Maybe we should remove all the rules beside dependencies? The current behavior might breaks validator with custom rules since the rules won't be removed from the queue.

funkyfuture: yeah, i thought about something like `Validator.rules_that_dont_validate_none_values` that can be set in custom validators like `Validator.priority_validations`.

PR #313, opened 2017-05-08
Fixes repeated normalization

funkyfuture:
i don't like it for its overhead, but it's the least 'hacky' compromise i could come up with.

cc @dkellner @mbarakaja

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
   - 3.5
   - 3.6
   - pypy
-  #- pypy3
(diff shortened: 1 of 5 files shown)


mbarakaja:
Sorry for not being in touch. I will give it a look and try this.

dkellner:
I think we have to live with this (small) overhead until we have a proper way of distinguishing between validation rules to be run before and after normalization. If they ever is a need for that apart from `readonly`.

PR #314, opened 2017-05-28
if a default_setter has a field argument pass the field name to that …

### LINKED ISSUE #279: [Feature Request] Default setter needs to know the field name
It would be very usefull to let the default setter know which field it is.
Use case:
- a filter can only have values which depends on the logged in user
- if the filter is missing it should default to all the values the user is allowed to pass in

Currently I work around this issue by having the default setter set it to `__all__` and then in the custom validator (which also validates the allowed values), and let it expand to actual default in `self.document`. Which might break in feature versions.

harvimt:
…argument

attempt to resolve #279

### DIFF
--- a/cerberus/validator.py
+++ b/cerberus/validator.py
@@ -785,12 +785,17 @@ class Validator(object):
                 {'type': 'callable'},
                 {'type': 'string'}
                 ]} """
+        import inspect
         if 'default_setter' in schema[field]:
             setter = schema[field]['default_setter']
             if isinstance(setter, _str_type):
                 setter = self.__get_rule_handler('normalize_default_setter',
                                                  setter)
-            mapping[field] = setter(mapping)
+            argspec = inspect.getfullargspec(setter)
+            if 'field' in argspec.args or 'field' in argspec.kwonlyargs:
+                mapping[field] = setter(mapping, field=field)
+            else:
+                mapping[field] = setter(mapping)
 
     # # Validating
 


funkyfuture:
thanks for addressing the issue. imo however, we need a generalized approach (alike [dependency_injection](https://dependency-injection-py.readthedocs.io)), that could then be re-used in several parts of the library's code and extensions.

if it's agreed to include such funtionality, it should be merged after a `1.2` release. it adds complexity, opens new opportunies and should thus be tested broadly by the community.

nicolaiarocci:
Yes, I'm going to release 1.2, then you guys can start working on this. Thanks @harvimt for starting work on the feature.

funkyfuture:
btw, i recently made use of such mechanics in [inxs](https://github.com/funkyfuture/inxs). it actually is pretty straight-forward and fun to use. however we must be aware that it adds some overhead. also, would we implement our own inspection or use the package mentioned above? i tend to the first option.

harvimt:
`inspect` is built into python. why would you "write your own inspection"?

On Tue, Jul 4, 2017 at 11:53 AM, Frank Sachsenheim <notifications@github.com

funkyfuture:
yes, `inspect` is a stdlib module. but comfortable [signature inspection](https://docs.python.org/3.7/library/inspect.html#introspecting-callables-with-the-signature-object) has been added with Python 3.3. as Python 2 is still supported, we'd have to implement that part too.

this could be an opportunity to start a discussion and layout a maintenance plan of future versions with regard to Python 2. in a seperate issue, @nicolaiarocci what do you think, shall i draft a rough sketch to start this?

nicolaiarocci:
> this could be an opportunity to start a discussion and layout a maintenance plan of future versions with regard to Python 2. in a seperate issue, @nicolaiarocci what do you think, shall I draft a rough sketch to start this?

Yes.

PR #317, opened 2017-06-19
Failing test: rules registry used in dict fields

nicolaiarocci:
It looks like rules added to the registry cannot be used within dict
fields. Would be nice to be able to use them in these scenarios, which
are pretty common with complex documents.


funkyfuture:
sorry, i don't have time for it atm, though it shouldn't be too difficult to solve. does the test fail due to a false result or due to an exception? in the latter case i might get a quick clue where the code should be poked.

nicolaiarocci:
Here is the stacktrace: https://travis-ci.org/pyeve/cerberus/jobs/244556168

funkyfuture:
d'oh, of course. but sorry, got no quick spot from that and some code reading.

funkyfuture:
i pledge to mark this for the `1.2` milestone.

Issue #299, opened 2017-02-23
NotImplementedError when accessing validator errors while using oneof list of dict


kardaj:
#### Use-case abstract

When I run the following snippet, using the `1.1` version, I get a `NotImplementedError`
```python
from cerberus import Validator
schema = {
    'x': {
        "type": "dict",
        "oneof": [{
            "schema": {
                'v': {'type': 'string'}
            }}, {
            "schema": {
                'v': {'type': 'list', 'schema': {'type': 'dict', 'schema': {
                    'a': {'type': 'string', 'required': True}
                }}},
            }}]
    }}

v = Validator(schema)
v.validate({'x': {'v': ['a']}})
v.errors
```

In this example, the problem is that `a` is supposed to be a `dict`. Instead, a `string` was found.

#### Support request

After looking into the code, I found out it's due to following section in the `insert_logic_error` method of the `BasicErrorHandler`.

```python
for child_error in child_errors:
    if child_error.is_logic_error:
        raise NotImplementedError
    elif child_error.is_group_error:
        raise NotImplementedError
```

Replacing it with the next snippet fixes my problem without breaking any of the existing tests of the project.
```python
for child_error in child_errors:
    if child_error.is_logic_error:
        self.insert_logic_error(child_error)
    elif child_error.is_group_error:
        self.insert_group_error(child_error)
```

But I guess this behavior is not an accident and there's a reason for the `NotImplementedError` being there. Am I missing something?

funkyfuture:
see #278

boramalper:
Same issue here! Can be reproduced using:

```import cerberus

generic_message_schema = {
    b"t": {"type": "binary", "empty": False, "required": True},
}

common_response_schema = {
    b"y": {"type": "binary", "allowed": [b"r"], "required": True}
}

common_response_arguments_schema = {
    b"id": {"type": "binary", "minlength": 20, "maxlength": 20, "required": True}
}

get_peers_response_schema = {
    **generic_message_schema,
    **common_response_schema,
    b"r": {"required": True, "schema": {
        **common_response_arguments_schema,
        b"token": {"type": "binary", "empty": False, "required": True},
        b"values": {"type": "list", "excludes": b"nodes", "required": True, "schema": {
            "type": "bytes",
            "minlength": 6, "maxlength": 6
        }},
        b"nodes": {"type": "binary", "excludes": b"values", "required": True, "minlength": 26,
                    "maxlength": 8 * 26}
    }}
}

[...]

Issue #281, opened 2016-11-11
docs: API reference at python-cerberus.org lacks generated contents


funkyfuture:
- I consulted these documentations:
  - [x] http://docs.python-cerberus.org/en/stable/
  - [x] http://docs.python-cerberus.org/en/latest/

- I consulted these sections of the docs:
  - API

- [x] I have the capacity to submit a patch when a bug is identified.

----

#### Bug report

i just noticed that these [generated](https://github.com/nicolaiarocci/cerberus/blob/master/docs/includes/generate.py) includes of the docs do not appear in the online docs while they do in a local build:

http://docs.python-cerberus.org/en/stable/api.html#error-codes
http://docs.python-cerberus.org/en/stable/api.html#schema-validation-schema

funkyfuture:
is the checkbox "Install your project inside a virtualenv using setup.py install" checked in the Advanced Settings of the RtD-project? if not, maybe the call to `docs/includes/generate.py` should be in `conf.py` rather than in the Makefile to let the build fail if there's a problem.

nicolaiarocci:
Yes the checkbox is checked now (it wasn't), but it does not seem to solve the issue.

funkyfuture:
now it also seems to affect [builds](https://travis-ci.org/pyeve/cerberus/jobs/249826748) on travis.

funkyfuture:
maybe a git-hook that updates the then repo-included assets is the most durable solution?

nicolaiarocci:
Seems odd to me that only py35 is affected on travis.

funkyfuture:
that is because the doctests run on it: https://github.com/pyeve/cerberus/blob/master/tox.ini#L36

funkyfuture:
i pledge to mark this for the `1.2` milestone.

PR #321, opened 2017-06-28
Add support for nested logic and group errors

### LINKED ISSUE #278: Seeing "TypeError: exceptions must derive from BaseException" when looking at validator.errors on failed validation
This issue tracker is dedicated to bugs and feature requests. 

Stack Overflow is the go-to place for general questions and general how-tos. On
SO, questions tagged with the 'cerberus' tag are actively monitored by the
project author, contributors and users.

When reporting a bug, please post an example and make it as simple as possible.
From your use-case throw everything out that is irrelevant for your question
/ isn't necessary to evoke the bug.

Try to adhere as much as possible to the template below:

Used Cerberus version / latest commit: ...
  - [x] http://docs.python-cerberus.org/en/stable/
  - [x] http://docs.python-cerberus.org/en/latest/
  - ...
  - ...
- [x] I found nothing relevant to my problem in the docs.
- [x] I found the documentation not helpful to my problem.

---
#### Use-case abstract

I am using a 'oneof_schema' rule in order to allow either:
- an empty dictionary, or
- a dictionary of a specific format

---

Using: Cerberus 1.0.1 and Python 3.5.2

I'm seeing a strange error when trying to use a 'oneof_schema' rule for my document. The goal is to either allow the 'measurement' field to be an empty dict, or following a specific format.

Examples of documents that (should) pass:

```
{
    'tokens': [
        {
            'measurement': {
                'qty': { 'num': 10, 'den': 1, },
                'units': 'lbs',
            },
        },
    ],
}

{
    'tokens': [
        {
            'measurement': {
                'qty': 100,
                'units': 'lbs',
            },
        },
    ],
}

{
    'tokens': [
        {
            'measurement': {},
        },
    ],
}
```

Example of a document that should fail:

```
{
    'tokens': [
        {
            'measurement': {
                'qty':     { 'num': 'THIS WILL FAIL!', 'den': 1, },
                'units':    'lbs',
            },
        },
    ],
}
```

Interestingly, when I tested the above "should-fail" document, I get a strange exception:

```
$ python failure.py
  File "main.py", line 64, in <module>
    print(v.errors)
  File "/home/cheffu/venvs/cerberus-cheffu-test/lib/python3.5/site-packages/cerberus/validator.py", line 360, in errors
    return self.error_handler(self._errors)
  File "/home/cheffu/venvs/cerberus-cheffu-test/lib/python3.5/site-packages/cerberus/errors.py", line 443, in __call__
    self.extend(errors)
  File "/home/cheffu/venvs/cerberus-cheffu-test/lib/python3.5/site-packages/cerberus/errors.py", line 372, in extend
    self.add(error)
  File "/home/cheffu/venvs/cerberus-cheffu-test/lib/python3.5/site-packages/cerberus/errors.py", line 453, in add
    self.insert_group_error(error)
  File "/home/cheffu/venvs/cerberus-cheffu-test/lib/python3.5/site-packages/cerberus/errors.py", line 496, in insert_group_error
    self.insert_logic_error(error)
  File "/home/cheffu/venvs/cerberus-cheffu-test/lib/python3.5/site-packages/cerberus/errors.py", line 516, in insert_logic_error
    raise NotImplemented
TypeError: exceptions must derive from BaseException
```

If I change the rule to remove the 'oneof_schema' rule (losing the ability to accept an empty dict), I'm able to see the validation error, as expected:

```
$ python ok.py
{'tokens': [{0: [{'measurement': [{'qty': [{'oneof': ['none or more than one rule validate', {'oneof definition 1': ['must be of integer type'], 'oneof definition 0': ['must be of integer type']}]}]}]}]}]}
```

The only difference in the code between the two files is the 'oneof_schema' rule in the 'measurement' schema (lines 36-39 in failure.py). Is this an error on my part, or something more insiduous?

Attached are the two Python files: [failure.py.txt](https://github.com/nicolaiarocci/cerberus/files/557555/failure.py.txt) and [ok.py.txt](https://github.com/nicolaiarocci/cerberus/files/557556/ok.py.txt).

kornholi:
Should fix #278 and #299. Not sure how I feel about modifying `document_path`, but I don't have the time right now to come up with something better.

### DIFF
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -476,6 +476,10 @@ class BasicErrorHandler(BaseErrorHandler):
 
     @encode_unicode
     def add(self, error):
+        # Make sure the original error is not altered with
+        # error paths specific to the handler.
+        error = deepcopy(error)
+
         self.rewrite_error_path(error)
 
         if error.is_logic_error:
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -476,6 +476,8 @@ class BasicErrorHandler(BaseErrorHandler):
 
     @encode_unicode
     def add(self, error):
+        self.rewrite_error_path(error)
+
         if error.is_logic_error:
             self.insert_logic_error(error)
         elif error.is_group_error:
@@ -530,24 +532,57 @@ class BasicErrorHandler(BaseErrorHandler):
                                   self.format_message(error.field, error))
 
     def insert_logic_error(self, error):
-        path = error.document_path + (error.rule, )
         field = error.field
+        self.insert_error(error.document_path,
+                          self.format_message(field, error))
 
-        self.insert_error(path, self.format_message(field, error))
-
-        for i in error.definitions_errors:
-            child_errors = error.definitions_errors[i]
+        for child_errors in error.definitions_errors.values():
             if not child_errors:
                 continue
-            nodename = '%s definition %s' % (error.rule, i)
+
             for child_error in child_errors:
                 if child_error.is_logic_error:
-                    raise NotImplementedError
+                    self.insert_logic_error(child_error)
                 elif child_error.is_group_error:
-                    raise NotImplementedError
+                    self.insert_group_error(child_error)
                 else:
-                    self.insert_error(path + (nodename,),
+                    self.insert_error(child_error.document_path,
                                       self.format_message(field, child_error))
+
+    def rewrite_error_path(self, error, offset=0):
+        """
+        Recursively rewrites the error path to correctly represent logic errors
+        """
+        if error.is_logic_error:
+            self.rewrite_logic_error_path(error, offset)
+        elif error.is_group_error:
+            self.rewrite_group_error_path(error, offset)
+
+    def rewrite_group_error_path(self, error, offset=0):
+        child_start = len(error.document_path) - offset
+
+        for child_error in error.child_errors:
+            relative_path = child_error.document_path[child_start:]
+            child_error.document_path = error.document_path + relative_path
+
+            self.rewrite_error_path(child_error, offset)
+
+    def rewrite_logic_error_path(self, error, offset=0):
+        error.document_path = error.document_path + (error.rule, )
+        child_start = len(error.document_path) - offset - 1
+
+        for i, child_errors in error.definitions_errors.items():
+            if not child_errors:
+                continue
+
+            nodename = '%s definition %s' % (error.rule, i)
+            path = error.document_path + (nodename,)
+
+            for child_error in child_errors:
+                rel_path = child_error.document_path[child_start:]
+                child_error.document_path = path + rel_path
+
+                self.rewrite_error_path(child_error, offset + 2)
 
     @property
     def pretty_tree(self):
--- a/cerberus/tests/test_validation.py
+++ b/cerberus/tests/test_validation.py
@@ -1222,8 +1222,8 @@ def test_anyof_schema(validator):
     scope = v_errors['parts'][-1][3][-1]['anyof'][-1]
     assert 'anyof definition 0' in scope
     assert 'anyof definition 1' in scope
-    assert scope['anyof definition 0'] == ["unknown field"]
-    assert scope['anyof definition 1'] == ["unknown field"]
+    assert scope['anyof definition 0'] == [{"product name": ["unknown field"]}]
+    assert scope['anyof definition 1'] == [{"product name": ["unknown field"]}]
     assert v_errors['parts'][-1][4] == ["must be of ['dict', 'string'] type"]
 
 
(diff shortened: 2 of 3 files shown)


cerberus/tests/test_validation.py
```diff
@@ -1276,6 +1276,39 @@ def test_nested_oneof_type():
...
+                'schema': {'bar': {'oneof_type': ['integer', 'float']}}
+            }},
+            {'baz': {'type': 'string'}}
+        ]}}
+
+    document = {'abc': {'foo': {'bar': 'bad'}}}
+
+    expected_errors = {
+        'abc': [
+            {'oneof': [
```
funkyfuture: do we really need this level?

cerberus/tests/test_validation.py
```diff
@@ -1276,6 +1276,39 @@ def test_nested_oneof_type():
...
+                'none or more than one rule validate',
+                {'oneof definition 0': [
+                    {'foo': [{'bar': [
+                        {'oneof': [
+                            'none or more than one rule validate',
```
funkyfuture: not your fault, but it should be 'validated'.

cerberus/errors.py
```diff
@@ -532,25 +534,58 @@ def insert_group_error(self, error):
...
+
+    def rewrite_logic_error_path(self, error, offset=0):
+        error.document_path = error.document_path + (error.rule, )
+        child_start = len(error.document_path) - offset - 1
+
+        for i, child_errors in error.definitions_errors.items():
+            if not child_errors:
+                continue
+
+            nodename = '%s definition %s' % (error.rule, i)
```
funkyfuture: i'm not sure whether there is a stringent way to format strings in the code base. if there's only or mostly `.format` used so far, please do the same.

funkyfuture:
i'm not getting the logs of the failed test against CPython 3.3. in doubt, i'm in favour of dropping its support.

cerberus/tests/test_validation.py
kornholi: all of the other errors are in present tense

cerberus/tests/test_validation.py
kornholi: line length

cerberus/errors.py
kornholi: from a quick `rg` it looks like an even mix

kornholi:
Not sure what you mean about changing bare errors. They should stay exactly the same, and none of the tests changed. Do you have an example?

I've been testing this with pretty complicated nested schemas and haven't hit any edge cases yet.

cerberus/tests/test_validation.py
funkyfuture: ah, i didn't mean indentation level in the code, but the level in the tree representation. the following two lines also mention the oneof rule. this isn't a request to remove it, but i want to raise the question.

funkyfuture:
interesting how fast the memory about one own's writing fades.. thanks for your checks on my misleading remarks.

when i'm not wrong:

- the handler iterates over the 'bare' error representations from `Validator._errors`
- these are used as sources to select and format the descriptional strings of the handler
- but in lines 568 and 573 you change the `document_path` attribute of the original error representation
- thus the original information is altered with a representational variation that is specific to the handler

the simplest solution is propably to add this method to `errors.ValidationError`:

```python
def __copy__
    return self.__class__(**vars(self))
```

and then use `copy.copy` to get an instance that can be manipulated. though this proposal isn't correct. the attributes `constraint`, `value` and `info` should be deepcopied before getting passed to the initialization to avoid problems with mutable values.

> I've been testing this with pretty complicated nested schemas and haven't hit any edge cases yet.

perfect. if these are sensible for you application, feel free to add reflecting regression tests.

cerberus/tests/test_validation.py
kornholi: yeah I was thinking the same thing. maybe change the children to just `definition n`?

funkyfuture: but then different of-rule-related errors (eg `oneof` and `anyof`) couldn't be discriminated.

kornholi: Right, I'm saying keep the `oneof` level but change `oneof definition n` to `definition n`.

kornholi:
Would [this deepcopy](https://github.com/pyeve/cerberus/pull/321/files#diff-b21a18a810079ce89a6e2d289ba98e93R481) be enough or am I missing something?

cerberus/tests/test_validation.py
funkyfuture: which will not work (and not communicate well if somehow mangled) if there are e.g. a `oneof` and an `anyof` rule for the same field and both yield errors.

funkyfuture:
i'm surprised, but it works:

```python
>>> from cerberus.errors import ValidationError
>>> from copy import deepcopy
>>> e = ValidationError((), (), 0, 'foo', {'x': ['a', 'b', 'c']}, None, ())
>>> f = deepcopy(e)
>>> f.constraint['x'].append('d')
>>> 'd' in e.constraint['x']
False
```

kornholi:
Is there anything left to do here? Not sure what's going on with the 3.5 tests.

cerberus/tests/test_validation.py
funkyfuture: this is still open, right?

funkyfuture:
> Not sure what's going on with the 3.5 tests.

this is #281

nicolaiarocci:
@kornholi thanks for your contribution!
@funkyfuture is this ready for merge? Let's ignore the failing doctest for the time being.

funkyfuture:
the redundant information and extra tree level of *of-errors is still [unsolved](https://github.com/pyeve/cerberus/pull/321#discussion_r124809777).

kornholi:
Could you provide a concrete example of how you want it to look? Unless I am misunderstanding you, it seems out of scope for this PR.

funkyfuture:
```python
expected_errors = {
    'abc': [
            ['none or more than one rule validate',
             {'oneof definition 0':
              [{'foo':
                [{'bar':
                 [{'oneof':
                  ['none or more than one rule validate',
                   {'oneof definition 0': ['must be of integer type'],
                    'oneof definition 1': ['must be of float type']}
                   ]}]}]}],
              'oneof definition 1': [{'foo': ['unknown field']}]}]
    ]
}
```

funkyfuture:
i pledge to mark this for the `1.2` milestone.

@kornholi maybe a refactoring of `Validator._drop_nodes_from_errorpaths`'s  mechanic would help with the superfluous docuemnt level.

nicolaiarocci:
Could you please rebase on top of master? Also it is my understanding that there's still some refactoring needed, see @funkyfuture comment above.

nicolaiarocci:
Rebased (once again) and merged. Thanks!

PR #332, opened 2017-08-15
Adding support for Decimal type (Decimal module)

Amedeo91:
This PR has the goal to provide support to validate decimal data type. 

This way allows checking whether the number in the request is parsable by the decimal module. Note: this is to provide support for the new decimal type in mongodb.

### DIFF
--- a/cerberus/tests/conftest.py
+++ b/cerberus/tests/conftest.py
@@ -62,6 +62,9 @@ sample_schema = {
         'min': 1,
         'max': 100,
     },
+    'a_decimal': {
+        'type': 'decimal',
+    },
     'a_set': {
         'type': 'set',
     },
--- a/cerberus/tests/test_validation.py
+++ b/cerberus/tests/test_validation.py
@@ -1,14 +1,18 @@
 # -*- coding: utf-8 -*-
 
+
+from bson import decimal128
+from datetime import datetime, date
+import decimal
+from random import choice
 import re
 import sys
-from datetime import datetime, date
-from random import choice
 from string import ascii_lowercase
 
 from pytest import mark
 
 from cerberus import errors, Validator
+from cerberus.platform import PYTHON_VERSION
 from cerberus.tests import \
     (assert_bad_type, assert_fail, assert_has_error, assert_not_has_error,
      assert_success, assert_document_error)
@@ -163,6 +167,14 @@ def test_not_a_binary():
 
 def test_not_a_integer():
     assert_bad_type('an_integer', 'integer', "i'm not an integer")
+
+
+def test_not_a_decimal():
+    assert_bad_type('a_decimal', 'decimal', "i'm not an decimal")
+    assert_bad_type('a_decimal', 'decimal', ".34.235.12")
+    assert_bad_type('a_decimal', 'decimal', "34.2")
+    assert_bad_type('a_decimal', 'decimal', 34.2)
+    assert_bad_type('a_decimal', 'decimal', 34)
 
 
 def test_not_a_boolean():
@@ -391,6 +403,18 @@ def test_string_allowed():
 
 def test_integer():
     assert_success({'an_integer': 50})
+
+
+def test_decimal():
+    # Module decimal does not accept integer or float for version 2.6
+    if PYTHON_VERSION != 2.6:
+        assert_success({'a_decimal': decimal.Decimal(50)})
+        assert_success({'a_decimal': decimal.Decimal(50.0)})
+    assert_success({'a_decimal': decimal.Decimal("50")})
+    assert_success({'a_decimal': decimal.Decimal("50.0")})
+    # I need to check this specificity for MongoDb data-type
+    assert_success({'a_decimal': decimal128.Decimal128("50")})
+    assert_success({'a_decimal': decimal128.Decimal128("50.0")})
 
 
 def test_boolean():
@@ -905,8 +929,8 @@ def test_self_root_document():
     class MyValidator(Validator):
         def _validate_root_doc(self, root_doc, field, value):
             """ {'type': 'boolean'} """
-            if ('sub' not in self.root_document or
-                    len(self.root_document['sub']) != 2):
+            if 'sub' not in self.root_document or \
+                    len(self.root_document['sub']) != 2:
                 self._error(field, 'self.context is not the root doc!')
 
     schema = {
@@ -960,15 +984,15 @@ def test_anyof():
     assert_success(doc, schema)
 
     # prop1 must be either a number between 0 and 10 or 100 and 110
-    schema = {'prop1': {'anyof':
-                        [{'min': 0, 'max': 10}, {'min': 100, 'max': 110}]}}
+    schema = {'prop1': {'anyof': [{'min': 0, 'max': 10},
+                                  {'min': 100, 'max': 110}]}}
     doc = {'prop1': 105}
 
     assert_success(doc, schema)
 
     # prop1 must be either a number between 0 and 10 or 100 and 110
-    schema = {'prop1': {'anyof':
-                        [{'min': 0, 'max': 10}, {'min': 100, 'max': 110}]}}
+    schema = {'prop1': {'anyof': [{'min': 0, 'max': 10},
+                                  {'min': 100, 'max': 110}]}}
     doc = {'prop1': 50}
 
     assert_fail(doc, schema)
--- a/cerberus/validator.py
+++ b/cerberus/validator.py
@@ -14,6 +14,7 @@ from ast import literal_eval
 from collections import Hashable, Iterable, Mapping, Sequence
 from copy import copy
 from datetime import date, datetime
+import decimal
 import re
 from warnings import warn
 
@@ -22,6 +23,14 @@ from cerberus.platform import _int_types, _str_type
 from cerberus.schema import (schema_registry, rules_set_registry,
                              DefinitionSchema, SchemaError)
 from cerberus.utils import drop_item_from_tuple, isclass
+
+DECIMAL_TYPES = (decimal.Decimal,)
+try:
+    from bson import decimal128
+except ImportError:
+    pass
+else:
+    DECIMAL_TYPES += (decimal128.Decimal128,)
 
 
 toy_error_handler = errors.ToyErrorHandler()
@@ -1216,6 +1225,10 @@ class Validator(object):
             self._error(field, errors.BAD_TYPE)
             self._drop_remaining_rules()
 
+    def _validate_type_binary(self, value):
+        if isinstance(value, (bytes, bytearray)):
+            return True
+
     def _validate_type_boolean(self, value):
         if isinstance(value, bool):
             return True
@@ -1228,6 +1241,10 @@ class Validator(object):
         if isinstance(value, datetime):
             return True
 
+    def _validate_type_decimal(self, value):
+        if isinstance(value, DECIMAL_TYPES):
+            return True
+
     def _validate_type_dict(self, value):
         if isinstance(value, Mapping):
             return True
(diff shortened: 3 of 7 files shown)


funkyfuture:
i'm not sure whether that makes sense. in opposite to other possible constraints, this test doesn't relate to an actual Python type. it tests whether a value can be used to initialize a `decimal.Decimal` object. that could also by achieved with `oneof`, `type` and `regex`. 

in the not-so-fundamental-question:

- there are no docs extension
- please keep methods sorted!
- version constants should be imported from the `platform` module

now, some more fundamental stuff again:

- the code shouldn't be available on platforms without the `decimal` module

i tend to recommend you to use it custom validators.

Amedeo91:
Hi!

decimal module is native in python since version 2.4, therefore all python version has it.

Regarding your question, I understand your point, but it is used a lot to provide support for prices and everywhere you want to keep the decimal part without approximation. 
Therefore, for my point of view, this validation should be native in the json validator. For the documentation, I will do it after I am able to make the CI passing.

Regards,
Amedeo

Amedeo91:
@nicolaiarocci I did the PR to start to add the decimal data type to Cerberus in order to have it available in pyeve/eve as well.

Regards,
Amedeo

Note: python 3.5 is failing for everybody for some problem in the documentation which i was not able to understand.

funkyfuture:
it'd make sense to test for `decimal.Decimal` after a value has been coerced. the way it is implemented now, it's just confusing because cerberus will tell you that something is of a typte that it isn't. this is everything else than intuitive with regards to the current semantics.

Amedeo91:
I do not really understand what it is not clear for you: how would you implement this behavior?

For my point of view, the Decimal module can be considered as a different way to represent numbers.
Therefore, it makes total sense for me to validate as a data type. But, if you have a different proposal, please let me know! :) 

Regards,
Amedeo

funkyfuture:
i have no doubt that `decimal.Decimal` representations are a valid representations of numbers.

> it makes total sense for me to validate as a data type.

yes, but then do exactly that. validate the type of a symbol, not what a value could be coerced to. if you want to use it as `Decimal`, you'd have to convert it in the normalization phase anyway.

> new decimal type in mongodb

just for my understanding, does the Python client have a strict concorance with the Python type of the same name?

Amedeo91:
Just for your understanding of the issue, look at this: https://github.com/pyeve/eve/issues/1045

App which uses mongodb needs to format the data type accordingly.

"yes, but then do exactly that. validate the type of a symbol, not what a value could be coerced to. if you want to use it as Decimal, you'd have to convert it in the normalization phase anyway." --> not able to follow you of how to do it: can you give me an example?

funkyfuture:
> not able to follow you of how to do it: can you give me an example?

the docs have [examples](http://docs.python-cerberus.org/en/stable/normalization-rules.html#value-coercion) on coercion.

from what i understand [here](https://api.mongodb.com/python/3.4.0/api/bson/decimal128.html), the MongoDB-client uses its own type.

Amedeo91:
decimal128 is a sub-class of the decimal module. The only difference is that it accepts only string as an input of the module, but I do not think we should enforce it in the json validator as it is a specificity of mongoDb.

For coerce, unfortunately, I need to use the type field in order to be able to implement it in Eve. For this point, I would prefer to wait for @nicolaiarocci feedback if you are ok with that so that we design together how to add the support in both Cerberus and Eve. Thank you for your feedback

Regards,
Amedeo

nicolaiarocci:
Since `decimal` is supported by Python, I think that supporting it as a base type is a good idea.

funkyfuture:
> Since decimal is supported by Python, I think that supporting it as a base type is a good idea.

there is no dissent in this regard.

though it becomes critical when more and more types from the standard library are supported and the footprint (import execution time, memory) of an application that uses cerberus increases with each imported module. so far only builtin types and some from `datetime` and `collections.abc` are imported.

> Special cases aren't special enough to break the rules.

my critique is still that the implementation makes no sense considering how Cerberus operates. it's understandable that the following inconsistency would alienate users:

```python
>>> schema = {'a_bool': {'type': 'boolean'},
...	      'a_date': {'type': 'date'},
...           'a_decimal': {'type': 'decimal'},
...           'another_decimal': {'type': 'decimal'}}
...
>>> document = {'a_bool': True,
...             'a_date': datetime.date(1945, 5, 9),
...             'a_decimal': 10,
...             'another_decimal': '100'}
...
>>> validator(document, schema)
True
>>> isinstance(document['a_bool'], bool)
True
>>> isinstance(document['a_date'], datetime.date)
True
>>> isinstance(document['a_decimal'], decimal.Decimal)
False
>>> isinstance(document['another_decimal'], decimal.Decimal)
False
```

> In the face of ambiguity, refuse the temptation to guess.

maybe a bad allegory, but who would trust a lying validator?

i'd appreciate feedback regarding the particular points i raise by anyone from the user community.

Amedeo91:
I understand youf point. What if we automatically convert the input to decimal in case it is parsable? After all, we are talking about number: I think it makes sense! Let me know your opinion. :)

Regards,
Amedeo

funkyfuture:
> What if we automatically convert the input to decimal in case it is parsable?

that is why i pointed to the value coercion. once a value is coerced it will pass a proper type test.

funkyfuture:
normalization is applied before validation. as documented.

Amedeo91:
Ok, I did the update in order to make it work as you requested. 

Regards,
Amedeo

funkyfuture:
propably the cleanest and safest way is to put this in the `platform` module:

```python
DECIMAL_TYPES = (decimal.Decimal,)
try:
    from bson import Decimal128
except ImportError:
    pass
else:
    DECIMAL_TYPES += Decimal128
```

and then use that constant as second argument for `isinstance`.

funkyfuture:
would it suffice to test that particular type in Eve as it already has that dependency?

cerberus/tests/test_validation.py
```diff
@@ -393,6 +404,22 @@ def test_integer():
     assert_success({'an_integer': 50})
 
 
+def test_decimal():
+    # Module decimal does not accept integer or float for version 2.6
+    if int(sys.version_info[0]) > 2 and int(sys.version_info[1]) > 6:
```
funkyfuture: there's the `PYTHON_VERSION` constant in `platform` for that.

cerberus/tests/test_validation.py
```diff
@@ -1270,8 +1297,20 @@ def test_oneof_schema():
 
 
 def test_nested_oneof_type():
-    schema = {'nested_oneof_type':
-              {'valueschema': {'oneof_type': ['string', 'integer']}}}
+    schema = \
```
funkyfuture: i appreciate the other reformattings, but this is unnecessary.

cerberus/validator.py
```diff
@@ -11,6 +11,7 @@
 from __future__ import absolute_import
 
 from ast import literal_eval
+import decimal
```
funkyfuture: please keep imports sorted.

cerberus/validator.py
```diff
@@ -1240,6 +1241,14 @@ def _validate_type_integer(self, value):
         if isinstance(value, _int_types):
             return True
 
+    def _validate_type_decimal(self, value):
+        if isinstance(value, decimal.Decimal):
```
funkyfuture: i'd prefer [this](https://github.com/pyeve/cerberus/pull/332#issuecomment-323454495).

docs/validation-rules.rst
```diff
@@ -648,6 +648,13 @@ Data type allowed for the key value. Can be one of the following names:
    * - ``string``
      - :func:`py2:basestring`
      - :class:`py3:str`
+   * - ``decimal``
+     - :class:`py2:decimal.Decimal()`
+     - :class:`py3:decimal.Decimal() or bson.decimal128.Decimal128()`
+
+.. versionchanged:: 1.1
```
funkyfuture: these comments are below.

cerberus/tests/test_validation.py
Amedeo91: Unfortunately, it is not imported in the test classes: what would you suggest?

cerberus/validator.py
Amedeo91: updating! Thanks

docs/validation-rules.rst
Amedeo91: updating

cerberus/validator.py
Amedeo91: updating

Amedeo91:
@funkyfuture I did the updated following your code review! :)

Thanks,
Amedeo

cerberus/tests/conftest.py
```diff
@@ -62,6 +63,13 @@ def validator():
         'min': 1,
         'max': 100,
     },
+    'a_decimal': {
+        'type': 'decimal',
+    },
+    'a_decimal_coercing': {
```
funkyfuture: this schema is only used for validation so far. it stems back from a time when Cerberus was far less complex. putting more complex stuff here has a chance to break a lot of tests.
there's a test module for normalizations you propably have noticed when you studied the code.

cerberus/tests/test_validation.py
```diff
@@ -1,14 +1,18 @@
 # -*- coding: utf-8 -*-
 
-import re
-import sys
+
+from bson import decimal128
```
funkyfuture: to repeat myself: would it suffice to test that in Eve? i'd rather avoid extra dependencies, even for testing.

Amedeo91: Sorry, I did not get this point: since it is a functionality offered by Cerberus, it should be tested within it. But I let you take the decision over it

cerberus/tests/conftest.py
Amedeo91: ok, I will find another place for it.

cerberus/validator.py
```diff
@@ -1240,6 +1249,11 @@ def _validate_type_integer(self, value):
         if isinstance(value, _int_types):
             return True
 
+    def _validate_type_decimal(self, value):
+        for decimal_type in DECIMAL_TYPES:
```
Amedeo91: oops! Sorry, I have it in my local! I will push now

Amedeo91:
@funkyfuture I do not really see why you are cynical in the first place.....
I am trying to follow your hints when I can and when I have time to do it

docs/validation-rules.rst
```diff
@@ -691,6 +695,9 @@ A list of types can be used to allow different values:
     fields. This allows to safely assume that field type is correct when other
     (standard or custom) rules are invoked.
 
+.. versionchanged:: 1.1
+  From this version, a new data type to support decimal module has been introduced.
```
Amedeo91: Sorry, I am updating the version and indentation :)

docs/validation-rules.rst
```diff
@@ -648,6 +648,10 @@ Data type allowed for the key value. Can be one of the following names:
    * - ``string``
      - :func:`py2:basestring`
      - :class:`py3:str`
+   * - ``decimal``
```
Amedeo91: ok

cerberus/validator.py
```diff
@@ -1240,6 +1249,11 @@ def _validate_type_integer(self, value):
         if isinstance(value, _int_types):
             return True
 
+    def _validate_type_decimal(self, value):
```
Amedeo91: They are already not sorted!! But ok, I will move mine

docs/validation-rules.rst
Amedeo91: The type is always available as decimal module. Only decimal128 might be not available as part of bson module (which exist for all python version)

cerberus/tests/test_validation.py
Amedeo91: put it back (I change it as I had a problem with the flake8 validation...but now it is working fine)

nicolaiarocci:
Decimal type support is going to be implemented in Eve's own custom validator instead.

Issue #333, opened 2017-08-18
docs build are failing due to an unknown cause


funkyfuture:
i just realized that the [failing doc build](https://travis-ci.org/pyeve/cerberus/jobs/265717785#L549) isn't caused by #281.

it started to fail on my local setup after i upgraded Sphinx. with 1.5.6 it's alright.

[this](https://github.com/celery/sphinx_celery/pull/2) raised my suspicion that it may be caused by [sphinxcontrib-issuetracker](https://pypi.python.org/pypi/sphinxcontrib-issuetracker) which hasn't been updated for over four years.

i opened https://github.com/ignatenkobrain/sphinxcontrib-issuetracker/issues/12.

funkyfuture:
there's a pull request pending, that should solve this issue: https://github.com/ignatenkobrain/sphinxcontrib-issuetracker/pull/13

nicolaiarocci:
Nice catch.

funkyfuture:
alternatively we could change to the [extlinks](http://www.sphinx-doc.org/en/stable/ext/extlinks.html) extension that is shipped with Sphinx and thus remove a dependency.

PR #334, opened 2017-08-18
Docs updates

funkyfuture:
this mainly addresses #281.

afaik RTD runs CPython 3.5, so this approach should work.

local builds with Sphinx 1.5.6 succeed as expected.


nicolaiarocci:
Looks good, but there is still a linkcheck failure preventing it to go all green on Py35

funkyfuture:
afaict the failure relates to #333. have you tried a local build with Sphinx 1.5 or w/o sphinxcontrib-issuetracker?

nicolaiarocci:
Right. Will do.

PR #335, opened 2017-08-20
Refactors type validation to a descriptive paradigm

funkyfuture:
this is not completed, just a proof-of-concept for a proposal.

as an aftermath of #332 i came to this approach to deal with type checks. it still supports custom 'type' testing methods (propably we all have misused them at some point), but i also propose to eventually remove this capability with the next breaking release for these reasons:

- just dealing with a mapping is obviously simpler than with methods
- it restricts to use the 'type' rule to really just test the type of an object
- and thus follows the recent development in Python to do better with types (type annotations, etc.)
- it also enforces (of course the appropriate 'vereindeutigen' seems to be a german-only term) the semantic of the 'type' rule's name

any other uses beside type checking that have been put in custom type validators can still be achieved with the 'validator' rule.

cc @Amedeo91

### DIFF
--- a/cerberus/tests/test_assorted.py
+++ b/cerberus/tests/test_assorted.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 
+from decimal import Decimal
+
 from pytest import mark
 
+from cerberus import TypeDefinition, Validator
 from cerberus.tests import assert_fail, assert_success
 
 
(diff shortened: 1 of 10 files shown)


cerberus/tests/test_assorted.py
```diff
@@ -27,3 +31,23 @@ def test_that_test_fails(test, document):
...
+    schema = {'measurement': {'type': 'decimal'}}
+
+    validator = Validator()
+    validator.types_mapping['decimal'] = decimal_type
+    # REMOVE
+    validator.rules['type']['allowed'] += ('decimal',)
+    assert_success(document, schema, validator)
+
+    class MyValidator(Validator):
+        types_mapping = Validator.types_mapping.copy()
```
funkyfuture: alternatively, the metaclass could do this and then update the copy with the `types_mapping` defined for the subclass. but this would make it impossible to populate that mapping from scratch.

nicolaiarocci:
I like it.

funkyfuture:
despite one minor flaw (see the `FIXME`) i finalized this.

funkyfuture:
hold on, there are some edges that i still need to address.

funkyfuture:
a'ight, it's ready for thorough reviews now.

nicolaiarocci:
Merged after rebase, see https://github.com/pyeve/cerberus/commit/3510b3cf631f580e91be7699c60ad93678a078ab

Thanks!

PR #340, opened 2017-09-26
Also consider the state of Validator.types_mapping in context of valid schemas

funkyfuture:
follow-up to #335. so changed `type_mappings` can not falsify `Validator._valid_schemas`.


nicolaiarocci:
rebased and merged

PR #370, opened 2018-03-28
docs updates

funkyfuture:
hopefully this will solve some doc issues. just ping me if there are issues after merging.

builds overview: https://readthedocs.org/projects/cerberus/builds/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,5 @@
+build:
+  image: latest
+python:
+  setup_py_install: true
+  version: 3.6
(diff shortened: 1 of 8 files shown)


nicolaiarocci:
looks like we aren't quite there yet https://readthedocs.org/projects/cerberus/builds/6961279/

funkyfuture:
a'ight, i added a fix for that in the same branch.

nicolaiarocci:
Works like a charm. A link is missing (or was it intentional) at the first 'external resource' reference. You made me laugh at the second one 😄

funkyfuture:
ah, thanks. pushed another commit with the link.

you know, everyone outside italy thinks that a 'cc'  is a hard k, while at the same time is able to properly pronounce capuccino. the trans-alpine barbarians need training on that issue.

PR #371, opened 2018-03-31
allow to set update=True  at instanciation

jgirardet:
This is to  have the same two instances but with different behaviour.
Needed if we can't access to validation parameter at validate call.

This PR keeps the standard behaviour unchanged

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,6 +29,7 @@ Patches and Suggestions
 - Gabriel Wainer
 - Harro van der Klauw
 - Jaroslav Semančík
+- Jimmy Girardet
 - Jonathan Huot
 - Kaleb Pomeroy
 - Kirill Pavlov
--- a/cerberus/tests/test_validation.py
+++ b/cerberus/tests/test_validation.py
@@ -55,6 +55,17 @@ def test_required_field(schema):
                        True))
 
 
+def test_update_at_instanciation(schema):
+    validator = Validator(schema=schema, update=True)
+    required_string_extension = {
+        'a_required_string': {'type': 'string',
+                              'minlength': 2,
+                              'maxlength': 10,
+                              'required': True}}
+    schema.update(required_string_extension)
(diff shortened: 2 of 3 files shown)


funkyfuture:
i think you could easily achieve that by subclassing `Validator` as there's a mechanism to store arbritrary initialization values:

```python

class MyValidator(cerberus.Validator):
    def validate(self, *args, **kwargs):
        kwargs['update'] = self._config['update']
        super().validate(*args, **kwargs)
```

jgirardet:
ok, that  solves the probleme.
however I think it's costless enhancement to add it by default, but maybe  there are some design choice I'm not aware.
Anyway thank you for your answer  and cerberus !

jgirardet:
Sorry but your proposal  doesn't work by me : 

``` python
import cerberus

class MyValidator(cerberus.Validator):
    def validate(self, *args, **kwargs):
        kwargs['update'] = self._config['update']
        super().validate(*args, **kwargs)


schema = {
    "a": {"type":"string", "required":True},
    "b": {"type":"string"},
}

val = MyValidator(schema)
valup = MyValidator(schema, update = True)

valid = {'a':'hello', 'b':"hello"}
invalid = {'b':"hello"}


print(val(valid))
print(val(invalid))
print(valup(valid))
print(valup(invalid))

"""
>>> True
>>> False
>>> True
>>> False      # Should be True, if working
"""
```

jgirardet:
ok after some tests, It doesn't work if y ou call validate or __call__ but it's ok with validated

jgirardet:
it's ok if I  don't use validate()  or __call_  but validated  :thinking:

jgirardet:
If you aren't intrested in it maybe we could close it.
I finally use  your subclass trick and call the data with validated() and it's good like that : 
https://github.com/jgirardet/apistar-cerberus/blob/master/apistar_cerberus/apistar_cerberus.py
thanks for herlp

PR #383, opened 2018-04-19
Various docs adjustments

funkyfuture:
how boring.


funkyfuture:
tests are falling because Python 3.3 is funky on Travis today and a linked article's website is currently in maintenance  mode, hence Sphinx' `linkcheck` fails.

nicolaiarocci:
Python 3.3, we could probably drop support for it. Thoughts?

funkyfuture:
upgrading code from 3.3 to 3.4 is only an effort in few cases. and who needs to do this, is overdue anyway.

we should look out for a service that provides more detailed stats about pypi-downloads for such questions.

on that occasion, would you propose black for Cerberus as well? now that quote-normalization can be skipped, i have no argument against it iirc.

funkyfuture:
that is what interestingly happens when `wheel` is about to be installed as dependency of `setuptools`:

    wheel requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*' but the running Python is 3.3.6

hence i conlcude that it should be dropped indeed, and 2.6 as well. that'd be parity with eve then.

nicolaiarocci:
The only concern I have is that dropping py26 and 33 is a breaking change. Since we're 1.x now, we should follow SemVer. 

One option is to simply skip py33 runs on Travis. Then, when we release 2.0, we'll officially drop both of them. Thoughts?

nicolaiarocci:
PS: @funkyfuture you should have write access to this repository. Use it to close/open and triage issues at will. Let's also both accept PRs before I merge them in. Cheers.

funkyfuture:
> The only concern I have is that dropping py26 and 33 is a breaking change. Since we're 1.x now, we should follow SemVer.

you have a point here logically, but i think it's not practical. in fact Cerberus has a depency for its installation: `setuptools`. and we can't control the version a user has installed. since at least its dependency  `wheel` doesn't support Python 3.3 anymore, we can't promise that Cerberus is even installable in the fashion we advice.

now, this doesn't mean that Cerberus isn't installable in another way and defunct.

two aspects have to be considered too: providing a test environment that works for all historical interpreters is an effort that must be justified. and very importantly, CPython has its very timed release schedules, we don't. given [this plan](https://devguide.python.org/#status-of-python-branches) and assuming that the PyPA people keep dropping support for an interpreter version around half a year of its eol (which, i'm really not sure, may be a statement that i read in the last months), Cerberus would have to release five major release within the next five years. and keeping that in sync with Cerberus-intrinsic deprecation cycles will be messy for maintainers and users.

in conclusion, i think we should clarify what our rules regarding semantic versioning are (omg, we should define an eol for every minor release, shouldn't we?) and that we reword from supported Python versions to tested ones. a commit w/ a proposal follows here.

> Let's also both accept PRs before I merge them in. Cheers.

perfect. i'm thinking about a bot that could label PRs with review-needed and ready-to-merge according to the review status. the maintenance of various version trees could maybe be eased with automatic reminders as well. haven't done any research on such yet and don't know when i could, but if you don't mind, could you create a repo `gh-bots` for idea discussion?

funkyfuture:
added mentioned commit.

PR #387, opened 2018-04-21
Adds the 'contains' rule

funkyfuture:
pretty straight-forward, isn't it?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,14 +22,17 @@ pip-log.txt
 .pytest_cache/
 .tox
 
-#Translations
+# Translations
 *.mo
 
-#Mr Developer
+# Mr Developer
 .mr.developer.cfg
 
-#Sphinx
+# Sphinx
 _build
+
+# Jetbrains
+.idea
 
 # my private cerberus playground
 run.py
--- a/AUTHORS
+++ b/AUTHORS
@@ -46,6 +46,7 @@ Contributors
 - Roman Redkovich
 - Sebastian Rajo
 - Sergey Leshchenko
+- Scott Crunkleton
 - Tobias Betz
 - Trong Hieu HA
 - Waldir Pimenta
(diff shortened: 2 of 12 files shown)


nicolaiarocci:
Screw it, since we were going to break semver with 1.2.x, we can as well go ahead and release the new stuff with that same version. Then we can go back to the regular workflow. Please let me merge until everything is sorted out.

nicolaiarocci:
rebased and merged

nicolaiarocci:
We might actually release this as 1.3 and be done with it :-)

funkyfuture:
> since we were going to break semver with 1.2.x

how were we? with #388? i'd consider documentation clarifications as bugfixes.

nicolaiarocci:
the drop of Python 2.6 and 3.3 support.

funkyfuture:
was that in the `1.2.x` branch? though for my understanding of [the release policy](https://github.com/pyeve/cerberus#versioning--interpreter-support) the drop of tested interpreters doesn't mean a minor branch has to be dropped. we 'just' can't test it. and we shouldn't release major releases whenever we can't test an interpreter anymore. as we still need to document git maintenance and release flows, we could think about including non-public, local tests based on Docker images into that.

now that this branch is deleted, we could still assemble a proper one.

nicolaiarocci:
Yes, it was in `1.2.x`. I agree that we can branch and cherry-pick useful stuff from master. Also, a `1.3` long-running branch might be doable too, as long as we keep it rebased on top of master, so it can be easily merged at due time (yes, we can force-push that one). I will try to find the time and write down the desired git workflow in the coming days.

funkyfuture:
i'll see that i make my [multi-interpreter environment](https://hub.docker.com/r/funkyfuture/nest-of-serpents/) for Docker fit for the old interpreters w/ old pip versions in the next days. but i wouldn't use that on Travis, that would be much heavier regarding needed resources.

nicolaiarocci:
I wouldn't worry too much about dropped pythons. Let's just wave goodbye to them 😄

Issue #384, opened 2018-04-21
Support for any iterable (instead of just list) in "allowed" rule.


crunk1:
Used Cerberus version / latest commit: 1.2
Python version: 3.6.5

- I consulted these documentations:
  - [x] http://docs.python-cerberus.org/en/stable/
  - [x] http://docs.python-cerberus.org/en/latest/

- [x] I found nothing relevant to my problem in the docs.
- [x] I found the documentation not helpful to my problem.

- [x] I have the capacity to improve the docs when my problem is solved.
- [x] I have the capacity to submit a patch when a bug is identified.

----

#### Use-case abstract

I suggest adding support for using iterables in the "allowed" validation rule. In particular, I have a string enum class derived from enum.Enum. An enum.Enum is iterable. I'll try to demonstrate my use case:
```python3
class MyEnum(StrEnum):
  FOO = 'FOO'

# Behavior of MyEnum
# list(MyEnum)         -> [<MyEnum.FOO: 'FOO'>]
# 'FOO' in MyEnum      -> True
# 'BAR' in MyEnum      -> False
# MyEnum.FOO in MyEnum -> True
# MyEnum.FOO == 'FOO'  -> True
# MyEnum.FOO == 'BAR'  -> False

# This works, but is undesirable:
schema = {
  'foo': {'type': 'list', 'allowed': list(MyEnum)},
}
v = cerberus.Validator(schema)
# v.validate({'foo': [MyEnum.FOO, 'FOO']}) -> True

# This raises an error:
schema = {
  'foo': {'type': 'list', 'allowed': MyEnum},
}
v = cerberus.Validator(schema)
# Traceback (most recent call last):
#   File "<input>", line 1, in <module>
#   File "C:\Python36\lib\site-packages\cerberus\validator.py", line 169, in __init__
#     self.schema = kwargs.get('schema', None)
#   File "C:\Python36\lib\site-packages\cerberus\validator.py", line 509, in schema
#     self._schema = DefinitionSchema(self, schema)
#   File "C:\Python36\lib\site-packages\cerberus\schema.py", line 69, in __init__
#     self.validate(schema)
#   File "C:\Python36\lib\site-packages\cerberus\schema.py", line 197, in validate
#     self._validate(schema)
#   File "C:\Python36\lib\site-packages\cerberus\schema.py", line 219, in _validate
#     raise SchemaError(self.schema_validator.errors)
[...]

PR #388, opened 2018-04-21
Change 'allowed' rule to use containers instead of lists.

crunk1:
Addresses #384.

### DIFF
--- a/docs/validation-rules.rst
+++ b/docs/validation-rules.rst
@@ -13,7 +13,7 @@ For a full elaboration refer to :ref:`this paragraph <allowing-the-unknown>`.
 
 allowed
 -------
-This rule takes a :term:`container` of allowed values.
+This rule takes a :class:`py3:collectionsabc.Container` of allowed values.
 Validates the target value if the value is in the allowed values.
 If the target value is an :term:`iterable`, all its members must be in the
 allowed values.


funkyfuture:
lptm.

funkyfuture:
you could actually add a test with your `Enum` use-case for Python 3.4+.

crunk1:
For simplicity, I don't think it is necessary to build out my entire Enum behavior. I'll keep it to testing a Container type other than list, ie I'll use a set.

By the way, I tried to modify the 'allowed' documentation, but I don't know where the Sphinx gloasary is. I wanted to add :term:`container`.

funkyfuture:
the [Python docs glossary](https://docs.python.org/3/glossary.html) does not include the term *container*. you may point to:

    :class:`py3:collectionsabc.Container`

crunk1:
Done.

docs/validation-rules.rst
```diff
@@ -13,7 +13,7 @@ For a full elaboration refer to :ref:`this paragraph <allowing-the-unknown>`.
 
 allowed
 -------
-This rule takes a :term:`container` of allowed values.
+This rule takes a :class:`py3:collectionsabc.Container` of allowed values.
```
funkyfuture: typo: `collections.abc`

Issue #390, opened 2018-04-26
A couple of spelling errors in the docs


ukch:
#### Support request / Bug report

When building the docs for distribution, I have run them through an automated spell-checker and come up with the following issues:

```
particulary :: particularly
allows to :: allows one to (*3)
occurance :: occurrence
intially :: initially
keywork :: keyword
superflous :: superfluous
lenght :: length
occured :: occurred
```
(In the table above, to the left is the actual spelling and to the right the expected spelling. `(*3)` means the issue occurs three separate times.)

Please consider correcting the spelling in the source files.

funkyfuture:
thanks for your efforts. as an open source community project we rely on patches from the communities.

nicolaiarocci:
Hello, thanks for the report. Would you mind submitting a pull request to fix them? We welcome first time contributors!

funkyfuture:
Resolved with #401.

PR #401, opened 2018-06-01
Doc misspells

Flargebla:
Just fixed some typos in the docs from Issue #390. The issue is a little old so not all of them still existed.

### DIFF
--- a/docs/errors.rst
+++ b/docs/errors.rst
@@ -58,7 +58,7 @@ instance after a processing of a document:
 
   - ``document_error_tree``: A ``dict``-like object that allows you to query
     nodes corresponding to your document.
-    The subscript notation on a node allows to fetch either a specific error
+    The subscript notation on a node allows you to fetch either a specific error
     that matches the given :class:`~cerberus.errors.ErrorDefinition` or a child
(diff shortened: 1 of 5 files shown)


docs/errors.rst
```diff
@@ -58,7 +58,7 @@ instance after a processing of a document:
 
   - ``document_error_tree``: A ``dict``-like object that allows you to query
     nodes corresponding to your document.
-    The subscript notation on a node allows to fetch either a specific error
+    The subscript notation on a node allows you to fetch either a specific error
```
funkyfuture: since there's no second singular around here and the docs usually (?) use `one` as in your other changes, it'd be good to apply that here too for stylistic consistency.

funkyfuture:
thanks.

funkyfuture:
can we document this in an extra repo as guideline for all repos in the `pyeve` namespace? we should also mind the maintenance of minor version branches .

nicolaiarocci:
probably a good idea, yes

funkyfuture:
as the `1.2.1` branch is published, there's no way to re-arrange that, right?

PR #407, opened 2018-06-07
Includes all error definition in the API docs

### LINKED ISSUE #404: Improve API docs on ErrorDefinitions
Used Cerberus version / latest commit: any

- [x] My question does not concern a practical use-case that I can't figure out 
      to solve.

----

#### Bug report

the api docs [exclude](https://github.com/pyeve/cerberus/blob/master/docs/includes/generate.py#L26) error definitions that are not assigned to a distinct rule. that's quark.

on that occasion, the semantics of the codes' bits should be documented better and bits that are safe to use for customizations should be declared (mind different notations!). ther might be an open issue related to the latter.

funkyfuture:
Also elaborates on the reserved and recommended bits of error codes.

Closes #404.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -53,7 +53,31 @@ Python Error Representations
 Error Codes
 ~~~~~~~~~~~
 
-These errors are used as :attr:`~cerberus.errors.ValidationError.code`.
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
rebased and merged

PR #419, opened 2018-07-05
Distinguished use of pre-commit for committing and linting

funkyfuture:
if i'm not wrong, malformatted code would have passed ci tests.

edit: i actually had tested that.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,8 +9,5 @@ repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v1.3.0
     hooks:
-    -   id: trailing-whitespace
-    -   id: end-of-file-fixer
-    -   id: check-yaml
-    -   id: debug-statements
-    -   id: flake8
+      - id: check-yaml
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -1,16 +1,19 @@
...
+      - id: black
+        args: [--quiet, --safe, --skip-string-normalization, cerberus, setup.py]
+        python_version: python3.6
+        types:
+          - python
+  - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v1.3.0
     hooks:
-    -   id: trailing-whitespace
-    -   id: end-of-file-fixer
```
nicolaiarocci: What is the reasoning behind removing `trailing-whitespace` and `end-of-file-fixer` hooks? Is Black going to take care of that (I admit I did not check the docs)

funkyfuture: yes, i checked both practically. otherwise black wouldn't be fun anyway.

nicolaiarocci: 👍

Issue #417, opened 2018-07-02
Proposal: Add an option to require all defined fields of a schema


pohmelie:
There are four cases which I think should be covered with validator instance arguments:
* Allow known missed, forbid unknown present (default behavior right now)
* Allow known missed, allow unknown present (`allow_unknown=True`)
* Forbid known missed, forbid unknown present (no way at validator creation time)
* Forbid known missed, allow unkown present (no way at validator creation time)

Cerberus have `allow_unknown` option for validator, but have no `allow_known_missed` / `forbid_known_missed` or something like this. As [stackoverflow](https://stackoverflow.com/questions/45935812/have-required-set-to-true-by-default-in-cerberus) question said, it is pretty annoying to set `required: True` for all fields, cause they can be nested and schema looks noisy and dirty.

pohmelie:
It looks like this can be achieved, by changin couple of places, where `'required'` checked to something like:
``` python
self.schema[field].get('required', self.forbid_known_missed)
```

pohmelie:
Ok, I have successfully patched cerberus `validator.py` with technic like above, tests are still ok and looks like added argument works fine. The only problem is since `'required'` became orphan there is an error [here](https://github.com/pyeve/cerberus/blob/master/cerberus/validator.py#L284), but I hope this can be resolved.

funkyfuture:
if this gets added, i'd prefer the option name `require_all`. some form of `required` is needed in the name to clearly relate it to the `required` rule.

should there also be a same-named flag for (sub-)schemas like `allow_unknown`?

funkyfuture:
@pohmelie a feedback on the progress would be great to have rather soonish as we're getting closer to the next minor release after which new features can't be merged for a while due to planned, larger code changes.

PR #421, opened 2018-07-07
add `require_all` validator argument

pohmelie:
#417


pohmelie:
The problem [place](https://github.com/pyeve/cerberus/pull/421/files#diff-795369a6cae83c62a70eafa3d3a1852eL284) resolved badly, since I'm not familiar with codebase. Need help here.

funkyfuture:
i don't have the time to look at it now, but i think a complementary rule like the `allow_unknown` rule is necessary because people will certainly have use-cases for this (where an overall option is too restrictive like `allow_unknown` was).

cerberus/validator.py
```diff
@@ -84,6 +84,9 @@ class BareValidator(object):
     :param allow_unknown: See :attr:`~cerberus.Validator.allow_unknown`.
                           Defaults to ``False``.
     :type allow_unknown: :class:`bool` or any :term:`mapping`
+    :param require_all: See :attr:`~cerberus.Validator.require_all`.
+                          Defaults to ``False``.
```
funkyfuture: this has a surplus of indentation.

cerberus/validator.py
```diff
@@ -281,7 +287,7 @@ def _error(self, *args):
                 if rule == 'nullable':
                     constraint = field_definitions.get(rule, False)
                 else:
-                    constraint = field_definitions[rule]
+                    constraint = field_definitions.get(rule)
```
funkyfuture: is this a special case for `required`? i'd rather keep that prone to a `KeyError` exception than have possibly incorrect `None` values in the data.

cerberus/validator.py
```diff
@@ -428,6 +435,19 @@ def allow_unknown(self, value):
...
+    @property
+    def require_all(self):
+        """ If ``True`` known fields that are defined in the schema will
+            be required.
+            Type: :class:`bool` """
+        return self._config.get('require_all', False)
+
+    @require_all.setter
+    def require_all(self, value):
+        if not (self.is_child or isinstance(value, (bool, DefinitionSchema))):
```
funkyfuture: you musn't copy that from `allow_unknown` where it has the purpose to validate the provided value.

cerberus/validator.py
pohmelie: Is it correct?
``` python
if rule == 'nullable':
    constraint = field_definitions.get(rule, False)
elif rule == 'required':
    constraint = field_definitions.get(rule, self.require_all)
else:
    constraint = field_definitions[rule]
```

cerberus/validator.py
pohmelie: `require_all` should not have setter? I'm not familiar with internals, sorry.

cerberus/validator.py
funkyfuture: looks good.

cerberus/validator.py
funkyfuture: it should have a setter. but the check in line 447-448 is specific to `allow_unknown`. just assign the value like `purge_readonly` for example.

cerberus/tests/test_validation.py
```diff
@@ -1830,3 +1830,50 @@ def test_contains(constraint):
         errors.MISSING_MEMBERS
     ].info[0]
     assert any(x in missing_actors for x in ('Eric Idle', 'Terry Gilliam'))
+
+
+def test_require_all_simple():
+    v = Validator({'foo': {'type': 'string'}}, require_all=True)
+    assert not v({})
```
funkyfuture: instead of assert statements you could use the helper functions [assert_fail](https://github.com/pyeve/cerberus/blob/master/cerberus/tests/__init__.py#L34) and [assert_success](https://github.com/pyeve/cerberus/blob/master/cerberus/tests/__init__.py#L72) that are more helpful for debugging. with the function that expects a failure one can also pass the expected error data, there are some uses of this as reference.

cerberus/tests/test_validation.py
```diff
@@ -1830,3 +1830,50 @@ def test_contains(constraint):
...
+    assert v({})
+    assert v({'foo': 'value'})
+    assert v({'bar': 'value'})
+    assert not v({'foo': 'value', 'bar': 'value'})
+
+
+def test_require_all_errors():
+    v = Validator({'foo': {'type': 'string'}}, require_all=True)
+    assert not v({})
+    assert v.errors == {'foo': ['required field']}
```
funkyfuture: a more detailed test (one should be enough) of the error data (all the properties of a [ValidationError](https://github.com/pyeve/cerberus/blob/master/cerberus/errors.py#L91)) with the functionality mentioned before would be preferable.

your test also checks the result of the default error handler, its behaviour could change and that wouldn't relate to your feature / these tests.

pohmelie:
@funkyfuture 
> __require_all__ similar to here

Sorry, have no idea what is going on here and how to make some similar changes to `__validate_required_fields` method. Need help.

cerberus/tests/test_validation.py
```diff
@@ -1838,3 +1838,71 @@ def test_contains(constraint):
...
+
+
+def test_require_all_simple():
+    schema = {'foo': {'type': 'string'}}
+    validator = Validator(require_all=True)
+    assert_fail(
+        {},
+        schema,
+        validator,
+        error=('foo', ('foo', 'required'), errors.REQUIRED_FIELD, True),
```
funkyfuture: as a start replace the second tuple value here with `'__require_all__'`. as far as i can oversee the situation, the `schema_path` variable in the `Validator._error` method must then be figured out with regards to the presence of the `required` rule for the submitted `field` in the schema when an `errors.REQUIRED_FIELD` is processed.

pohmelie: That is obvious... and this is test, I mean, at first we need to fix code, so it will result `__require_all__` in error schema path and only after tests should be fixed. The problem is to fix `validator.py` and replace `required` with `__require_all__` in `__validate_requried_fields` in some cases.

funkyfuture: > and this is test, I mean, at first we need to fix code

following the paradigm of test-driven development, one first formulates assertions in a test.

`__validate_required_fields` calls `_error` which only takes a reference to the field and an error definition as argument, hence the detemination of the correct (pseudo-) schema path has to happen in the latter method.

pohmelie: > following the paradigm of test-driven development, one first formulates assertions in a test.

Ok, but this does not help me at all. TDD is for developers, this PR won't be accepted without tests and implementation, so there is no reason to order changes.
> hence the detemination of the correct (pseudo-) schema path has to happen in the latter method.

I don't get it. We have `__validate_unknown_fields`, which calls `_error` with some preparation and we have `__validate_required_fields`, which calls `_error` without something. Why this cases are not equal? What is "later method" in this case?

funkyfuture: i don't get it.

> TDD is for developers, this PR won't be accepted without tests and implementation, so there is no reason to order changes.

afaict you are developing a feature here, hence you're in the role of a developer. i don't care how you work if the result is fine from a variety of angles, one of these are tests that define expected behaviour and are passing. at the moment the test defines an undesired behaviour.

> Why this cases are not equal?

`__validate_unknown_fields` uses a 'child validator' that handles the schema path (that may eventually be used in `_error`).

> `__validate_required_fields`, which calls `_error` without something

no. that never happens.

> What is "later method" in this case?

the latter that was mentioned in the previous part in the sentence, `_error`.

pohmelie: > if the result is fine from a variety of angles, one of these are tests that define expected behaviour and are passing. at the moment the test defines an undesired behaviour.

That is the point. I understand what behaviour we need, but I don't understand how to do it. And advice to "start with test" sound like a joke. :smile: 

> the latter that was mentioned in the previous part in the sentence, `_error`.

You suggest to add `__require_all__` to `schema_path` in `_error` method, am I right?

funkyfuture: > And advice to "start with test" sound like a joke.

neither did i intend to be funny with that statement nor can i find traces of humour in it.

so far, my successes with that approach were really fun, but not laughable.

> You suggest to add `__require_all__` to schema_path in `_error` method, am I right?

more specificly: `__require_all__` *is* the `schema_path` when the error was defined by the `require_all` option of the root validator.

funkyfuture:
a hint on a not so obvious spot: the third paragraph in `schema.rst` also needs to be updated.

pohmelie:
@funkyfuture 
[This](https://github.com/pyeve/cerberus/blob/master/docs/schemas.rst#validation) paragraph? How this should be updated?

pohmelie:
> on that occasion: what's your overall status on this feature?
Sorry, I'm not native english speaker, so I don't get you... As I understand you, then answer is: «I can't check readiness of this feature, since I'm contributor and you are reviewer. It's up to you do decide, that my implementation is good enough or not».

funkyfuture:
i'm not a native english speaker as well (okay, i have an advantage w/ my mother tongue) and even as a citizen of the gdr i didn't learn russian. it seems we have no other choice than using the international language of science and technology. maybe you can get support with that.

regarding the documentation i can provide corrected formulations when reviewing.

so far the changes look good, but the whole feature is still lacking these to major aspects:

- [ ] an `require_all` rule that can override the configuration value of an instance when a subdocument is validated with `schema`
- [ ] an enhancement of the documentation in all appropriate places

again, the `allow_unknown` rule can give good orientation on both, but it's not identical.

docs/schemas.rst
```diff
@@ -15,7 +15,8 @@ which is expected to be a string not longer than 10 characters. Something like
 very long string'}`` or ``{'name': 99}`` would not.
 
 By default all keys in a document are optional unless the :ref:`required`-rule
-is set for a key.
+is set for a key or validator attribute :attr:`~cerberus.Validator.require_all`
```
funkyfuture: "is set ``True`` for individual fields or the validator's :attr:`~cerberus.Validator.require_all` is set to ``True`` in order to expect all schema-defined fields to be present in the document.

docs/usage.rst
```diff
@@ -162,6 +162,67 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+-------------
+By default any keys defined in the schema are not required:
+
+.. doctest::
+
+    >>> schema = {'name': {'type': 'string', 'maxlength': 10}}
+    >>> v.validate({}, schema)
+    True
+
+However, you can require all document keys pairs by either setting
```
funkyfuture: if there's an 'either', what's the other case?

docs/usage.rst
```diff
@@ -162,6 +162,67 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
    ``allow_unknown`` can also be set to a validation schema.
 
 
+.. _requiring-all:
+
+Requiring all
```
funkyfuture: this looks like copied from the `allow_unknown` section. given that `require_all` is less complex, a sufficient explanation should be doable with less prosa.

docs/validation-rules.rst
```diff
@@ -660,6 +660,16 @@ library's :mod:`re`-module.
...
+-----------
+This can be used in conjunction with the  `schema <schema_dict-rule>`_ rule
+when validating a mapping in order to set the
+:attr:`~cerberus.Validator.require_all` property of the validator for the
+subdocument.
+For a full elaboration refer to :ref:`this paragraph <requiring-all>`.
+
 .. _required:
 
 required
```
funkyfuture: please add a short cross-reference to the section in `usage.rst` here.

docs/usage.rst
```diff
@@ -162,6 +162,67 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+    ...     }
+    ...   }
+    ... }
+
+    >>> v.validate({'a_dict': {}}, schema)
+    False
+
+    >>> v.validate({'a_dict': {'address': 'foobar'}}, schema)
+    True
+
```
funkyfuture: please add a short cross-reference to the `require` and `require_all` rules here.

cerberus/validator.py
```diff
@@ -1043,6 +1072,7 @@ def validate_rule(rule):
                        {'type': ['dict', 'string'],
                         'check_with': 'bulk_schema'}]} """
     )
+    _validate_require_all = dummy_for_rule_validation("{'type': 'boolean'}")
```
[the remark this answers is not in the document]
pohmelie: it looks like they are not in alphabetical order right now:
```
logical
anyof
nullable
keysrules
```

docs/usage.rst
pohmelie: Have no idea how to reduce text. Exclude some examples?

cerberus/tests/test_validation.py
```diff
@@ -5,6 +5,7 @@
 from datetime import datetime, date
 from random import choice
 from string import ascii_lowercase
+import itertools
```
funkyfuture: please shift this above `re`.

docs/usage.rst
```diff
@@ -162,6 +162,69 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
 
+.. _requiring-all:
+
+Requiring all
+-------------
+See also :ref:`this paragraph <require_all>`, :ref:`this paragraph <required>`.
+
+By default any keys defined in the schema are not required:
+
+.. doctest::
```
funkyfuture: i don't think this example is needed.

docs/usage.rst
```diff
@@ -162,6 +162,69 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+
+    >>> schema = {'name': {'type': 'string', 'maxlength': 10}}
+    >>> v.validate({}, schema)
+    True
+
+However, you can require all document keys pairs by setting
```
funkyfuture: i think that suffices as a sentence after the explanation of the variant when it is set at initialization. an inline example (`validator_require_all = True`) should also be enough, no need for a code block.

docs/usage.rst
```diff
@@ -162,6 +162,69 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+    False
+
+``require_all`` can also be set at initialization:
+
+.. doctest::
+
+    >>> v = Validator({'name': {'type': 'string', 'maxlength': 10}}, require_all=True)
+    >>> v.validate({})
+    False
+    >>> v.require_all = False
```
funkyfuture: actually, this shoull be enough with regards to the validator attribute.

docs/usage.rst
```diff
@@ -162,6 +162,69 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+    >>> v.validate({})
+    True
+
+``require_all`` can also be set as rule to configure a validator for a nested
```
funkyfuture: i'm not sure, but i think we commonly use the term subdocument instead of 'nested mapping' through the docs.

docs/usage.rst
```diff
@@ -162,6 +162,69 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+
+    >>> v = Validator()
+    >>> v.require_all
+    False
+
+    >>> schema = {
+    ...   'name': {'type': 'string'},
+    ...   'a_dict': {
+    ...     'type': 'dict',
+    ...     'require_all': True, # this overrides the behaviour for
```
funkyfuture: the comments' content has already been stated.

docs/usage.rst
```diff
@@ -162,6 +162,69 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+    ...     'schema': {          # the validation of this definition
+    ...       'address': {'type': 'string'}
+    ...     }
+    ...   }
+    ... }
+
+    >>> v.validate({'a_dict': {}}, schema)
```
funkyfuture: the example would make more sense if there was another field defined in the schema at the top level, that doesn't exist in the document. the output of `v.errors` could conclude the story.

docs/usage.rst
```diff
@@ -162,6 +162,69 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+    False
+
+    >>> v.validate({'a_dict': {'address': 'foobar'}}, schema)
```
funkyfuture: not really needed imo.

docs/validation-rules.rst
```diff
@@ -660,10 +660,22 @@ library's :mod:`re`-module.
...
 --------
+See also :ref:`this paragraph <requiring-all>`.
```
funkyfuture: rather put this in a first note admonition below, a little more elaborated: "To define all fields of a document ..."

funkyfuture:
what about this?

```python
schema = {
    'foo': {
        'type': 'dict',
        'require_all': True,
        'allow_unknown': {…},
    }
}
```

would that make something explode? should `require_all` and `allow_unknown` mutually exclude each other?

pohmelie:
> would that make something explode?

At first look they are not crossing anywhere.

funkyfuture:
> At first look they are not crossing anywhere.

should the child validator that is dispatched for `allow_unknown` have the `require_all` argument always set ro `False`?

docs/usage.rst
```diff
@@ -168,37 +168,12 @@ Requiring all
...
-    >>> v.validate({})
-    False
-    >>> v.require_all = False
-    >>> v.validate({})
-    True
-
-``require_all`` can also be set as rule to configure a validator for a nested
-mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
+``require_all`` to ``True`` at validator initialization (``v = Validator(…, require_all=True)``)
+or change latter via attribute access (``v.require_all = True``).
```
funkyfuture: … change it later …

pohmelie:
> should the child validator that is dispatched for `allow_unknown` have the `require_all` argument always set ro `False`?

Hm, if it validates only unknown fields, then «yes», I think.

pohmelie:
It looks like squash broke all merge logic, since all history goes away. And I don't known how to squash only my commits, cause there was merges from master to my branch between. You can squash all my pull to one commit on merge actually, and squashed changes are on "Files" tab above.

Anyway, do not how to restore this PR to previous state. :disappointed:

funkyfuture:
sorry, i don't know what the situation in your local repository actually looks like. i hope you have all your changes still on your machine. just apply them on a new branch checked out from master an open a new pr.

PR #425, opened 2018-07-19
Omits warning when no schema for the meta rule constraint is avail

funkyfuture:
As per default there is none defined and that gets noisy.


nicolaiarocci:
Rebased and merged

PR #426, opened 2018-07-19
Adjustments for pre-commit

funkyfuture:
i was wrong about black's capabilities or rather i missed the point that flake8 does additionaly detect unused imports. also i missed the point that pre-commit calls the plugins per changed file.

i'm not sure if i got it correctly. when black reformats code, this change doesn't get added to the commit and i have to commit a second time. is this supposedly so?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,11 @@
-.*
+.benchmarks
+.cache
+.dockerignore
+.eggs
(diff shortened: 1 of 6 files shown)


nicolaiarocci:
> is this supposedly so?

yep! rebased and merged.

Issue #385, opened 2018-04-21
Cerberus 2: Proposal to clearly distinguish names for different 'schema' rules


funkyfuture:
There are currently five rules to define constraints on container objects that are either sequences or mappings:

- `items` defines a sequence of individual sets of rules for each item in a sequence
- `keyschema` defines one set of rules that all keys of a mapping must match
- `schema`
  - either defines a schema for a subdocument (a mapping) with the same (slightly enhanced, actually) semantics as a 'top level' schema
  - or it defines one set of rules that all items in a sequence must match
- `valueschema` defines one set of rules that all values of a mapping must match

#### Proposal 1

the different meanings of `schema` have been a source of confusion among users and it also forces us to operate with some assumptions in the code and to add some necessary overhead around this.

therefore the next major relase should clarify semantics on the `schema` rule. it's quiet obvious that its name points to the first mentioned mapping-related semantic as it matches the overall usage of the term *schema* in Cerberus' nomenclature.

the name of the other rule should be chosen wisely to imply both the similarity with and the difference to the `items` rule.

i can think of the following atm:

* ~~`allitems`~~ (see rationale in P2)
* ~~`eachitem`~~ (see rationale in P2)
* `itemsrules`

#### Proposal 2

as the comprehension above shows, there are two categories of rules in this realm: those that define a set of rules for particular items or just any item of a container object and those that define a schema that assigns such set of rules to values identified by a key. it'd be therefore terminologically stringent to rename `keyschema` and `valueschema` to `keysrules` and `valuesrules`. as there are no name conflicts, the previous names can still be used as aliases.

with these terms, `itemsrules` would be a consistent choice for the previous proposal as the morpheme `items` implies that it concerns all of the sequence members and the morpheme `rules` implies that a set of rules is expected.

funkyfuture:
note: from an implementation perspective `itemrules` and `keyrules` might actually be aliases in the end as both validate the values that an iteration over a container object yields against the given set of rules.

funkyfuture:
i'd say we include proposal 2 in the 1.3 release with the same rationale [as here](https://github.com/pyeve/cerberus/issues/405#issuecomment-398759449).

nicolaiarocci:
Addressed with #422

funkyfuture:
implemented in dca25495bf5cf55dbecd2bbbf436752d63735d79

Issue #374, opened 2018-04-05
Cerberus 2: Proposal for stricter and more flexible type checks at the same time!


funkyfuture:
i propose to make the type checking mechanics simpler and more flexible regarding the possible constraints with the next major release of Cerberus.

this means that users need to refactor their custom validators - remove any method prefixed with `_validate_type_` - resp. the employed schemas when upgrading. uses with the vanilla validator may require schema changes due to slight semantical shifts. type checks can no longer be misused for what rules and validators are supposed for.

users could publish enhanced sets of type definitions, e.g. for uses with NumPy or other common frameworks that define data types.

an implementation would roughly look like this (in the style proposed in #372):

```python

class Validator(metaclass=ValidatorMeta):
    type_definitions = {
        'binary':
            TypeDefinition('binary', (collections.abc.ByteString), ()),
        'boolean':
            TypeDefinition('boolean', (bool,), ()),
        'bytearray':  # new to enhance builtins coverage
            TypeDefinition('bytearray', (bytearray,), ()),
        'bytes':  # new to enhance builtins coverage
            TypeDefinition('bytes', (bytes,), ()),
        'complex':  # new to enhance builtins coverage
            TypeDefinition('complex', (complex,), ()),
        'date':
            TypeDefinition('date', (datetime.date,), ()),
        'datetime':
            TypeDefinition('datetime', (datetime.datetime,), ()),
        'dict':  # this is stricter again, but unambigious
            TypeDefinition('dict', (dict,), ()),
        'float':  # stricter, but unambigious
            TypeDefinition('float', (float,), ()),
        'frozenset':  # new to enhance builtins coverage
            TypeDefinition('frozenset', (frozenset,), ()),
        'integer':
            TypeDefinition('integer', (int,), ()),
        'list':  # this is stricter again, but unambigious
            TypeDefinition('list', (list,), ()),
        'mapping':  # new
            TypeDefinition('mapping', (collections.abc.Mapping,), ()),
        'none':   # does this obsolete the 'null' rule?
            TypeDefinition('none', (type(None),), ()),
        'number':
            TypeDefinition('number', (int, float), (bool,)),
        'sequence':  # new; should AnyStr and ByteString be excluded?
            TypeDefinition('sequence', (collections.abc.Sequence), ()),
        'set':
            TypeDefinition('set', (set,), ()),
        'string':  # this includes bytes again
            TypeDefinition('string', (typing.AnyStr,), ()),
[...]

PR #431, opened 2018-08-04
A preview on Cerberus 2

funkyfuture:
i thought it might be a good idea to present what i have done so far for Cerberus 2.

any questions and critique is welcome, reviewing per commit is certainly to be recommended.

**don't merge this, it is wip, it will be rewritten**


.travis.yml
```diff
@@ -2,11 +2,8 @@ sudo: false
 language: python
 cache: pip
 python:
-  - 2.7
-  - 3.4
   - 3.5
   - 3.6
-  - pypy
```
jeking3: Add 3.7 so you get all the deprecation warnings for 3.8

cerberus/__init__.py
```diff
@@ -8,22 +8,68 @@
 
 """
 
-from __future__ import absolute_import
+from typing import Dict, Optional, Tuple, Union
```
jeking3: In a project locally I added a new rule called "content:" that takes types like "uuid" and the rule maps their processing to python builtins from a string.  It's something I think is missing when using cerberus as a yaml import validation where everything is a string type (and thus you cannot use custom types).

Another thing cerberus is missing is conditional logic, for example it would be nice to be able to say something required only if another field is set to "X".  For example: "required: true if type == 'number'".

UPGRADING.rst
```diff
@@ -1,5 +1,57 @@
...
+Rules
+.....
+
+A few rules have been renamed:
+
+========================  ===============
+Old name                  New name
+========================  ===============
+``keyschema``             ``keysrules``
+``schema`` for sequences  ``itemsrules``
```
jeking3: `schema` was a better choice for this, IMO.

funkyfuture: please discuss this in #385. in any case: we need to clearly distinguish the currently two `schema` rules.

.travis.yml
funkyfuture: done: 28e1ee75d3c03024964c45347cc3bf04a8db5c9a

cerberus/__init__.py
funkyfuture: regarding types, please see #374. there are the *of-rules for conditional logic. they are causing enough trouble, everything that needs to act more distinguished must be solved with customizations.

eirnym:
Could you fix merge conflicts?

funkyfuture:
@eirnym this will happen once 1.3 is released (and thus the feature freeze is in effect) and i find the time to rebase or to rewrite, it's not a trivial task. also, i want to open a discussion on the question of the extent of type annotations.

what is your particular interest to ask for an update? maybe there's another way to serve that.

eirnym:
I'd like to help you with as many changes there make sense. Also I'd like to see a few additional features asked here and there like "user documentation" for each element inside the schema itself and checking a schema object.

Currently, I remove `doc` keys to pass them forward and checking a schema object by checking a dummy document.  

Also I want to try to find out if it's hard to merge the new version without rebasing and breaking anything.

eirnym:
I opened a pull request to your branch with the merge, please, verify it.

funkyfuture:
> Also I'd like to see a few additional features asked here and there like "user documentation" for each element inside the schema itself and checking a schema object.

please open separate issues for each proposal.

> Also I want to try to find out if it's hard to merge the new version without rebasing and breaking anything.

i certainly won't hinder you from giving it a shot.

having that said, i don't think the type annotations will sustain in the extent as i did here. but they can also be (partly) removed later.

> This library is quite sensitive for my work, so I have a motivation to support it.

everyone is welcome. i'd have to review all the planning to give some practical advice what should be addressed in what order. maybe you could look through the items in the `ROADMAP.md` and post in the corresponing issues that you would like to work on.

funkyfuture:
Closing in favor of #497.

PR #434, opened 2018-08-13
add `require_all` (close #417)

### LINKED ISSUE #417: Proposal: Add an option to require all defined fields of a schema

pohmelie:
#417 
#421

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -171,8 +171,9 @@ By default any keys defined in the schema are not required.
 However, you can require all document keys pairs by setting
 ``require_all`` to ``True`` at validator initialization (``v = Validator(…, require_all=True)``)
 or change it latter via attribute access (``v.require_all = True``).
-``require_all`` can also be set as rule to configure a validator for a subdocument
-that is checked against the :ref:`schema <schema_dict-rule>` rule:
+``require_all`` can also be set :ref:`as rule <require_all>` to configure a
+validator for a subdocument that is checked against the
+:ref:`schema <schema_dict-rule>` rule:
 
 .. doctest::
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -162,6 +162,46 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
    ``allow_unknown`` can also be set to a validation schema.
 
 
+.. _requiring-all:
+
+Requiring all
+-------------
+See also :ref:`this paragraph <require_all>`, :ref:`this paragraph <required>`.
+
+By default any keys defined in the schema are not required.
+However, you can require all document keys pairs by setting
+``require_all`` to ``True`` at validator initialization (``v = Validator(…, require_all=True)``)
+or change it latter via attribute access (``v.require_all = True``).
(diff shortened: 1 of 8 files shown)


docs/usage.rst
```diff
@@ -162,6 +162,46 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
    ``allow_unknown`` can also be set to a validation schema.
 
 
+.. _requiring-all:
+
+Requiring all
+-------------
+See also :ref:`this paragraph <require_all>`, :ref:`this paragraph <required>`.
```
funkyfuture: i don't think this is needed...

docs/usage.rst
```diff
@@ -162,6 +162,46 @@ mapping that is checked against the :ref:`schema <schema_dict-rule>` rule:
...
+
+By default any keys defined in the schema are not required.
+However, you can require all document keys pairs by setting
+``require_all`` to ``True`` at validator initialization (``v = Validator(…, require_all=True)``)
+or change it latter via attribute access (``v.require_all = True``).
+``require_all`` can also be set as rule to configure a validator for a subdocument
```
funkyfuture: when `require_all` (the rule) is linked here.

docs/validation-rules.rst
```diff
@@ -679,6 +690,10 @@ unless :meth:`~cerberus.Validator.validate` is called with ``update=True``:
     >>> v.validate(document, update=True)
     True
 
+.. note::
+
+   To define all fields of a document as required see :ref:`this paragraph <requiring-all>`.
```
funkyfuture: 'paragraph' -> 'section about the available options' !?

docs/usage.rst
pohmelie: don't get how this should be changed.
```
``require_all`` can also be set as rule to configure a validator for a subdocument
that is checked against the :ref:`schema <schema_dict-rule>` rule:
```
→
```
``require_all`` can also be set as rule to configure a validator for a subdocument
when ``require_all`` (the rule) is linked here.
```
?

funkyfuture: i mean this:

```
``require_all`` can also be set :ref:`as rule <require_all>` to configure …
```

funkyfuture:
thanks a lot!

pohmelie:
@funkyfuture thanks for help! :wink:

nicolaiarocci:
thanks and sorry for the delay

Issue #427, opened 2018-07-26
'coerce' still ignores 'nullable': True


andreymal:
#269 was fixed, but only for integer:

```
>>> v = cerberus.Validator({'foo': {'coerce': int, 'nullable': True, 'type': 'integer'}})
>>> v.normalized({'foo': None})
{'foo': None}
```

Just replace integer to string:

```
>>> v = cerberus.Validator({'foo': {'coerce': str, 'nullable': True, 'type': 'string'}})
>>> v.normalized({'foo': None})
```

Expected behavior: `{'foo': None}`

Actual behavior: `{'foo': 'None'}`

So I still have to use a workaround `'coerce': lambda x: str(x) if x is not None else None`

But now I'm not sure if it's a bug or a feature (duplicate of #142?). If this is a feature, please describe this behavior [in the documentation](http://docs.python-cerberus.org/en/latest/normalization-rules.html#value-coercion)

funkyfuture:
this behaviour is to be expected as `str(None)` is not failing. i'd say a custom coercer is the way to go.

would *not coercing* values that are `None` and constrained with `nullable: True` be the better tactic?

funkyfuture:
sorry for not being very verbose on this. i gave it some thoughts:

the behaviour you describe is to be expected because the current implementation works like this:
when a coercion fails on a field that has `nullable` set to `True` and the coercion failed due to a `TypeError` (like `int(None)` would raise), that failure is ignored - no error is logged, the input value is kept in the document. there's actually no test whether the input value of the failed coercion is `None`.

to comply with what i understand as your requirement, it would be simpler and clearer: when a field has
the value `None` and the schema allows this with a positively defined `nullable` the value would not get coerced at all.

my concern with that second approach is, that there would be no way to coerce a `None` value of a field that is eventually allowed to be `None`. that may be the case when the coercion routine depends on the state of other data (document contents or configuration) which may or may not result in another value than `None`.

a good counter-argument against my concern would be, that such resolutions should be handled with a `default_setter`. the problem is, that these are never applied to fields that have that positively defined `nullable`.

i haven't looked into the implications if that ignorance would be changed, but my guts tell me that this would be the clearest concept:
- `default_setter`s don't ignore `nullable` fields
- but `coerce` does

does this make sense? would that break current uses?

andreymal:
>there would be no way to coerce a None value of a field that is eventually allowed to be None

Good argument

[...]

PR #435, opened 2018-08-15
Fix test default none nonnullable

dkellner:
see #427

```
Fix `test_{default,default_setter}_none_nonnullable`

Both tests called the wrong helper function.

Furthermore `_test_default_none_nonnullable` mimicked the behaviour of
`_test_default_existent`, now it actually tests what its name suggests.
```

cc @funkyfuture


nicolaiarocci:
rebased and merged, thanks.

PR #445, opened 2018-09-16
Feature: Document level validations

### LINKED ISSUE #444: Proposal:  Add document level validation
#### Use-case abstract
Sometimes a document should be considered invalid based on the values of multiple fields.  For example, in a medical record where  `gender==male` and `is_pregnant==True` is not valid, even though each field looked at individually complies with the schema (`'allowed': ['male', 'female']` and `'type': 'boolean'` respectively). 

----

#### Feature request

Some documents as a whole should be considered invalid based on more than one field.  Take this schema, for example:
```
schema = {
    'name': {'type': 'string'},
    'aka': {'type': 'string'},
    'gender': {
        'type': 'string',
        'allowed': ['male', 'female']
    },
    'is_pregnant': {'type': 'boolean'}
}
```
This document should be invalid because males can't be pregnant (even if they have the right to be):
```
medical_record = {
    'name': 'Stan',
    'aka': 'Loretta',
    'gender': 'male',
    'is_pregnant': True
}
```
(with apologies to [Life of Brian](https://www.youtube.com/watch?v=sFBOQzSk14c))

This can be accomplished currently with a custom validator like this:
```
class MyValidator(Validator):
    def _validate_preg_check(self, preg_check, field, value):
        """ {'type': 'boolean'} """
        if field == 'gender':
            gender = value
            is_pregnant = self.document.get('is_pregnant')
        elif field == 'is_pregnant':
            gender = self.document.get('gender')
            is_pregnant = value
        else:
            return

        is_valid = True
        if is_pregnant:
            is_valid = gender == 'female'
        if not(preg_check and is_valid):
            self._error(field, 'Invalid: males cannot be pregnant')
```
Now we add the rules to the schema:
```
schema = {
    'name': {'type': 'string'},
    'aka': {'type': 'string'},
    'gender': {
        'type': 'string',
        'allowed': ['male', 'female'],
        'preg_check': True
    },
    'is_pregnant': {
      'type': 'boolean', 
      'preg_check': True
    }
}
```
and validation fails as we expect...
```
validator = MyValidator(schema)
validator.validate(medical_record)
```
...with the following errors dict:
```
{
  'gender': ['Invalid: males cannot be pregnant'], 
  'is_pregnant': ['Invalid: males cannot be pregnant']
}
```

What's wrong with this?
- have to apply the rule to each field involved in the check
- therefore, the rule fires for each field when it needs to only fire once
- the custom validator has to have an `if/elif/else` to cover all fields involved
- the errors dict doesn't cleanly tell the story.  There is nothing wrong with either the gender field nor the is_pregnant field.  The only problem is those two values together in the same document

This example involves only two fields.  The problems compound when there are many more fields and many more document level checks.

Better is a simpler custom validator that looks like this (with the `if/elif/else` gone):
```
class MyValidator(Validator):
    def _validate_document_preg_check(self, preg_check, field, value):
        """ {'type': 'boolean'} """
        is_valid = True
        if self.document.get('is_pregnant'):
            is_valid = self.document.get('gender') == 'female'
        if not(preg_check and is_valid):
            self._error(field, 'This medical record is invalid')
```
Now the schema can return to the original:
```
schema = {
    'name': {'type': 'string'},
    'aka': {'type': 'string'},
    'gender': {
        'type': 'string',
        'allowed': ['male', 'female']
    },
    'is_pregnant': {'type': 'boolean'}
}
```
This validation also fails as we expect...
```
validator = MyValidator(schema, document_validations={'preg_check': True})
validator.validate(medical_record)
```
...but this time with the following errors dict:
```
{
  '_document': ['Invalid: males cannot be pregnant']
}
```

biscuit314:
Closes #444 - see issue for details.  Causes this test to pass:

```
def test_document_level_validation():
    class MyValidator(Validator):
        def _validate_document_preg_check(self, preg_check, field, value):
            """ {'type': 'boolean'} """
            is_valid = True
            if self.document.get('is_pregnant'):
                is_valid = self.document.get('gender') == 'female'
            if not(preg_check and is_valid):
                self._error(field, 'Invalid: males cannot be pregnant')

    schema = {
        'name': {'type': 'string'},
        'aka': {'type': 'string'},
        'gender': {
            'type': 'string',
            'allowed': ['male', 'female']
        },
        'is_pregnant': {'type': 'boolean'}
    }

    medical_record = {
        'name': 'Stan',
        'aka': 'Loretta',
        'gender': 'male',
        'is_pregnant': True
    }
    validator = MyValidator(schema, document_validations={'preg_check': True})
    assert_fail(medical_record, validator=validator)
```


funkyfuture:
this might be an interesting, slim approach to deal w/ an often encountered kind of problems. to review it however, a pr that is narrowed to it is necessary.

PR #448, opened 2018-09-24
Honor `ignore_none_values` for excludes

NotSpecial:
The `ignore_none_values` allows to treat None like a missing field, i.e.
ignore fields set to None.
This should also extend to the `exludes` validator: Fields set to None should
be ignored and not raise an error, if `ignore_none_values` is True.

Contributions:

- Include check for `if_none_values` in `_validate_excludes`

- Test the feature

- Updates to AUTHORS and CHANGES

### DIFF
--- a/cerberus/validator.py
+++ b/cerberus/validator.py
@@ -1171,7 +1171,14 @@ class BareValidator(object):
 
                 self._unrequired_by_excludes.add(excluded_field)
 
-        if any(excluded_field in self.document for excluded_field in excluded_fields):
+        def _ignore_field(field):
+            # If the field is None and ignore_none_values is set, ignore it.
+            return self.document[field] is None and self.ignore_none_values
+
+        if any(
+            excluded_field in self.document and not _ignore_field(excluded_field)
+            for excluded_field in excluded_fields
+        ):
             exclusion_str = ', '.join(
                 "'{0}'".format(field) for field in excluded_fields
             )
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,6 +10,7 @@ Core maintainers
 Contributors
 ~~~~~~~~~~~~
 
+- Alexander Dietmüller
 - Antoine Lubineau
 - Arsh Singh
 - Audric Schiltknecht
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,7 @@ New
   the next major release of Cerbers (`#385`_)
 - The ``meta`` pseudo-rule can be used to store arbitrary application data
   related to a field in a schema
+- The `excludes` validator takes `ignore_none_values` into account
 - **Python 2.6 and 3.3 are no longer supported**
 
 Fixed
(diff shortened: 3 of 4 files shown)


cerberus/validator.py
```diff
@@ -1171,7 +1171,14 @@ def _validate_excludes(self, excluded_fields, field, value):
 
                 self._unrequired_by_excludes.add(excluded_field)
 
-        if any(excluded_field in self.document for excluded_field in excluded_fields):
+        def _ignore_field(field):
+            # If the field is None and ignore_none_values is set, ignore it.
+            return self.document[field] is None and self.ignore_none_values
+
+        if any(
+            excluded_field in self.document and not _ignore_field(excluded_field)
```
funkyfuture: this function call in a loop is suboptimal performance-wise, and i see not much gain in expressiveness here.

funkyfuture:
maybe this makes more understandable what i mean: to extend the semantics of `ignore_none_values` to referenced fields would mean to also change the `dependencies` rule. possibly more.

NotSpecial:
Thank you for your quick response. You raise important points regarding `dependencies`.

The starting point of this work was the documentation, which states that `ignore_none_values` controls whether [None values in the document are processed or not](http://docs.python-cerberus.org/en/stable/api.html?highlight=ignore%20none#cerberus.Validator.ignore_none_values).

As I interpret this, setting `ignore_none_values` to true should tell Cerberus to really ignore all None values.

Currently, the behaviour is a bit inconsistent. None values are ignored for `required` validators, yet processed for `excludes`, `dependencies` (and any others?)

We require an option to really ignore `None` values and consistently treat them the same as missing fields, as in our usecase a distinction between a missing field and a field set to None does not make sense and would only cause confusion.

Do you think there would be a better way to approach this? Can you think of a case where only a partial ignoring of None is useful?

If you think this is a useful way to go in general, I'll gladly extend the PR to include `dependencies` as well (and other rules, if I can find any).

funkyfuture:
i can see how the documentation on that option is certainly not telling the complete story about its implementation / effects. but i would consider this a documenation bug not one of the implementation.

for context, the initial addition of the option actually only targeted type checks (96532fc8efbc0b057dd6cd23d0324c8c5a929456) and the got broadened to all rules.

> We require an option to really ignore None values and consistently treat them the same as missing fields, as in our usecase a distinction between a missing field and a field set to None does not make sense and would only cause confusion.

you may use the normalization phase to remove these fields from a document - though with the available rules that may not be straightforward, overriding the `_normalize_mapping` method should facilitate that very simply. or you throw your documents on a similarly simple function before passing them to the validator.

> Can you think of a case where only a partial ignoring of None is useful?

no and yes, by engaging phantasy i might come up w/ something. but experience shows that someone already did and will open a bug report after such change. at least chances are good.

funkyfuture:
@NotSpecial did you investigate to solve your problem in normalization phase? are there any issues left beside the too unspecific documentation?

funkyfuture:
closing this until further feedback.

Issue #436, opened 2018-08-22
Forbidden rule does not support float


MartyHub:
The following schema does not work as expected :

```json
{
  "amount": {
    "forbidden": [0],
    "max": 1,
    "min": 0,
    "type": "number",
  }
}
```

because [the forbidden rule](https://github.com/pyeve/cerberus/blob/7d61ac3624d34d5dcb621089e48d91a3d1ee64a7/cerberus/validator.py#L1189) only check for `int` :

```python
Validator().validate(
  {
    "amount": 0.0
  },
  schema
)
```

funkyfuture:
yeah, that last clause should probably be `else` and strings should be tested there too.

safaozturk93:
I'll look at this.

PR #449, opened 2018-09-30
fix(validator): Forbidden rule does not support float

safaozturk93:
#436

### DIFF
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -70,7 +70,7 @@ SEQUENCE_SCHEMA = ErrorDefinition(0x82, 'schema')
 # TODO remove KEYSCHEMA AND VALUESCHEMA with next major release
 KEYSRULES = KEYSCHEMA = ErrorDefinition(0x83, 'keysrules')
 VALUESRULES = VALUESCHEMA = ErrorDefinition(0x84, 'valuesrules')
-BAD_ITEMS = ErrorDefinition(0x8f, 'items')
+BAD_ITEMS = ErrorDefinition(0x8F, 'items')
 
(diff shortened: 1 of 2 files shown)


cerberus/errors.py
```diff
@@ -70,7 +70,7 @@
 # TODO remove KEYSCHEMA AND VALUESCHEMA with next major release
 KEYSRULES = KEYSCHEMA = ErrorDefinition(0x83, 'keysrules')
 VALUESRULES = VALUESCHEMA = ErrorDefinition(0x84, 'valuesrules')
-BAD_ITEMS = ErrorDefinition(0x8f, 'items')
+BAD_ITEMS = ErrorDefinition(0x8F, 'items')
```
funkyfuture: that is an unnecessary change.

cerberus/validator.py
```diff
@@ -1189,6 +1189,11 @@ def _validate_forbidden(self, forbidden_values, field, value):
         elif isinstance(value, int):
             if value in forbidden_values:
                 self._error(field, errors.FORBIDDEN_VALUE, value)
+        elif isinstance(value, float):
```
funkyfuture: please keep it simple. afaict, there's no need to branch off for float or integer values, only for container values that are not strings.

cerberus/errors.py
safaozturk93: this violates black reformatting rule in travis pipeline.

cerberus/validator.py
safaozturk93: I will look at it.

funkyfuture:
Closing in favour of #455.

PR #464, opened 2019-01-28
Adapts Travis' config to run on Xenial

funkyfuture:


funkyfuture:
i have no clue yet why the obsolete interpreters don't show up yet.

funkyfuture:
the state of current affairs is that the generated test matrix is as expected and that the interpreters from the `deadsnakes` repo are installed. `tox`, however uses `python2.7` for `py26`, `py31` to `py33`.

does anyone have an idea what's missing?

nicolaiarocci:
Looks like we have a winner 👏

funkyfuture:
> Looks like we have a winner

because everything is green? unfortunatley not, see [this issue](https://github.com/tox-dev/tox/issues/1168). i'm already thinking about dumping `tox`.

funkyfuture:
okay, while i'm testing another framework to organize the various tests, i realize that even old, available pytest versions often won't run on obsolete interpreters or the Cerberus (test) code uses language features that leads to failure. so, i get CPython 3.3 from the outdated interpreters to successfully run the test suite and that's it.

so, even using simple, reliable tools and scripts won't yield useful results; hence we can stick with `tox`, but i don't see that we can provide information regarding obsolete interpreters even in the short term.

PR #472, opened 2019-03-12
Catch exception if the distribution is not found

### LINKED ISSUE #471: Set __version__ to an invalid value if package distribution is not found
Used Cerberus version / latest commit: master

- [x] My question does not concern a practical use-case that I can't figure out
      to solve.

----

#### Use-case abstract

When attempting to use the cerberus repo as a git submodule in a project, users are going to get a `DistributionNotFound` error, because the [get_distribution()](https://github.com/pyeve/cerberus/blob/f4b78f8a9002b6464eaf1a5ee714986621be3497/cerberus/__init__.py) method in the `__init__.py` file cannot retrieve the distribution of the package.

See stack trace:

```
...
    from cerberus import Validator
  File "/Users/s4heid/workspace/some-project/lib/cerberus/cerberus/__init__.py", line 24, in <module>
    __version__ = pkg_resources.get_distribution('Cerberus').version
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 481, in get_distribution
    dist = get_provider(dist)
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 357, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 786, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'Cerberus' distribution was not found and is required by the application
```

I currently workaround this issue by manually setting `__version__` to some string.

----

#### Bug report / Feature request

I would expect that if `pkg_resources.DistributionNotFound` is raised, the exception is getting catched and `__version__` set to an invalid version string which won't break your source installation and at the same time indicate that something is broken if the distribution cannot be found.

The  name of the version string could be something like `unknown`, which I've seen in other projects as well.

s4heid:
When the distribution of Cerberus cannot be found, catch the exception
instead of raising an error and exiting the program. This allows to
consume the package without installing it, e.g. as a git submodule. In
the exception case the __version__ string is set to `unknown`.

Fixes #471


cerberus/__init__.py
```diff
@@ -10,14 +10,17 @@
...
+from pkg_resources import get_distribution, DistributionNotFound
 
 from cerberus.validator import DocumentError, Validator
 from cerberus.schema import rules_set_registry, schema_registry, SchemaError
 from cerberus.utils import TypeDefinition
 
 
-__version__ = pkg_resources.get_distribution('Cerberus').version
+try:
+    __version__ = get_distribution(__name__).version
```
funkyfuture: I think using `__name__` here is too dynamic and it would break if this part is moved to anotehr module.

funkyfuture:
sorry, i forgot that the code to be tested it executed at module evaluation. the unit-testable approach would be to put the questionable in a function and call that from the module body. i'd say do as you like.

cerberus/__init__.py
s4heid: I was wondering whether it could/should be more generic, because I didn't find a test which covers the case you describe. But I see your point and change it back again.

s4heid:
Agreed; I also had this in mind at the beginning. Another solution would be to mock the `pkg_resources.get_distribution` method and patch the builtin __import__ function. wdyt?

cerberus/tests/conftest.py
```diff
@@ -1,12 +1,49 @@
...
 
 from copy import deepcopy
+from pkg_resources import Distribution, DistributionNotFound
 
 import pytest
 
 from cerberus import Validator
 
 
+def reload_module(name):
```
funkyfuture: though this is no runtime code, i'd still prefer to rely on `cerberus.platform.PYTHON_VERSION` instead of a nested try-except-block.

cerberus/tests/test_init.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
```
funkyfuture: i'd say the `test_assorted` is still roomy enough to contain these tests.

cerberus/tests/conftest.py
```diff
@@ -1,12 +1,49 @@
...
+            from imp import reload
+    reload(name)
+
+
+@pytest.fixture
+def cerberus():
+    return __import__('cerberus')
+
+
+@pytest.fixture(scope="function")
```
funkyfuture: as these fixtures aren't reused, i think the monkeypatching should happen in the test functions to make it more obvious what it is about and what's happening. is the `cerberus` fixture even necessary with that style?

cerberus/tests/test_init.py
```diff
@@ -0,0 +1,13 @@
...
+
+import cerberus
+
+
+def test_pkgresources_version(get_distribution_finds_distribution):
+    version = cerberus.__version__
```
funkyfuture: an expression in one line would be fine.

cerberus/tests/conftest.py
s4heid: I guess you are right and we should monkeypatch inside the test function. Might be clearer to read.

cerberus/tests/conftest.py
s4heid: sure, that might be nicer! using `reload` across multiple python versions is a pain.

s4heid:
Thanks for the review, @funkyfuture. Your suggestions made perfect sense to me.

funkyfuture:
thanks a lot for your contribution!

PR #490, opened 2019-05-02
Don't ignore all exceptions during coercions for nullable fields

nikhaldi:
Only ignore exceptions if a value for a nullable field is not null.
Previously all exceptions during custom coerction were ignored with nullable
fields. The developer and/or user most likely wants to know about
exceptions that happen during coercion - this leads to hard to debug
problems otherwise.

The purpose of the `e is not TypeError` condition was unclear. As far as I
know this condition can never be false (`e` is an instance of an exception,
not an exception class).

I think this is still in line with the original intent of this code, which
was to make common coercions like `float` work with nullable fields out
of the box: https://github.com/pyeve/cerberus/issues/269

### DIFF
--- a/cerberus/tests/test_normalization.py
+++ b/cerberus/tests/test_normalization.py
@@ -170,6 +170,18 @@ def test_nullables_dont_fail_coerce():
     assert_normalized(document, document, schema)
 
 
+def test_nullables_fail_coerce_on_non_null_values(validator):
+    def failing_coercion(value):
+        raise Exception("expected to fail")
+
+    schema = {'foo': {'coerce': failing_coercion, 'nullable': True, 'type': 'integer'}}
+    document = {'foo': None}
+    assert_normalized(document, document, schema)
+
+    validator({'foo': 2}, schema)
+    assert errors.COERCION_FAILED in validator._errors
+
+
 def test_normalized():
     schema = {'amount': {'coerce': int}}
     document = {'amount': '2'}
--- a/cerberus/validator.py
+++ b/cerberus/validator.py
@@ -723,7 +723,7 @@ class BareValidator(object):
         try:
             return processor(value)
         except Exception as e:
-            if not nullable and e is not TypeError:
+            if not (nullable and value is None):
                 self._error(field, error, str(e))
             return value
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -42,6 +42,7 @@ Contributors
 - Martin Ortbauer
 - Matthew Ellison
 - Michael Klich
+- Nik Haldimann
 - Nikita Melentev
 - Nikita Vlaznev
 - Paul Weaver
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,9 @@ Improved
 ~~~~~~~~
 - Add ``maintainer`` and ``maintainer_email`` to setup.py (`#481`_)
 - Add ``project_urls`` to setup.py (`#480`_)
+- Don't ignore all exceptions during coercions for nullable fields. If a
+coercion raises an exception for a nullable field where the field is not
+``None`` the validation now fails.
 
 .. _`#484`: https://github.com/pyeve/cerberus/issues/484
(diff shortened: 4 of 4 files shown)


nikhaldi:
The Travis failure for `test_nested_error_paths ` is only on Python 3.5. Unclear to me what the problem is. It doesn't reproduce for me locally on 3.5. I'm going go force push a noop change to get Travis to rerun.

nikhaldi:
The same test passed on the second try, so just appears to be flaky. The failure was https://travis-ci.org/pyeve/cerberus/jobs/527412260

cerberus/tests/test_normalization.py
```diff
@@ -170,6 +170,15 @@ def test_nullables_dont_fail_coerce():
     assert_normalized(document, document, schema)
 
 
+def test_nullables_fail_coerce_on_non_null_values(validator):
+    schema = {'foo': {'coerce': lambda x: 1 / 0, 'nullable': True, 'type': 'integer'}}
```
funkyfuture: a coercion function that explicitly raised an exception with a simple message like "expected to fail" would make the intent clearer.

cerberus/tests/test_normalization.py
```diff
@@ -170,6 +170,15 @@ def test_nullables_dont_fail_coerce():
...
+    document = {'foo': None}
```
funkyfuture: this is already covered in the previous test.

cerberus/validator.py
```diff
@@ -723,7 +723,7 @@ def __normalize_coerce(self, processor, field, value, nullable, error):
         try:
             return processor(value)
         except Exception as e:
-            if not nullable and e is not TypeError:
+            if not nullable or value is not None:
```
funkyfuture: i find this term more obvious: `not (nullable and value is None)`

funkyfuture:
> The same test passed on the second try, so just appears to be flaky.

yes, Python 2.7 and 3.5 sort errors kind of erraticly. no worries.

cerberus/validator.py
funkyfuture: second thought: should the coercion get tried at all when the part in brackets is true?

cerberus/tests/test_normalization.py
nikhaldi: Done

cerberus/tests/test_normalization.py
nikhaldi: True, but I feel the repetition is a good idea here because the new behavior is about the difference between these two cases (the value is `None` or not).

cerberus/validator.py
nikhaldi: I changed the condition.

Not sure there's an obvious answer to whether the coercion should be tried or not. Somebody might want to coerce `None` to something else? Or would the validation in that case fail at a later stage anyway?

nikhaldi: I just realized that NOT trying the coercion for nullable fields when the value is None would fix this issue: https://github.com/pyeve/cerberus/issues/427

To be honest, I don't feel I'm in a position to make the right call here. So happy to make this change or not.

nikhaldi:
Side note, `test_nested_error_paths` has failed 4 out of 5 times on Travis for me now. I'm not going to force a Travis rerun again until I know this is ready to merge.

cerberus/validator.py
funkyfuture: thanks for digging this out. that issue concluded to postpone changes until the 2.0 release; and i think i shall document that this still needs to be layed out design-wise.

funkyfuture:
> has failed 4 out of 5 times on Travis for me now

now, that's bad luck. but just ignore it.

cerberus/tests/test_normalization.py
funkyfuture: i'm not saying you're wrong, but we don't do that elsewhere (oh dear, formulating this makes me feel old). i'm gonna amend this when merging.

Issue #515, opened 2019-11-18
conditional normalization using oneof


linupi:
Used Cerberus version / latest commit: current master

- [x] I have the capacity to improve the docs when my problem is solved.
- [x] I have the capacity to submit a patch when a bug is identified.
- [x] My question does not concern a practical use-case that I can't figure out
      to solve.

----

#### Use-case abstract

So far Cerberus does not allow for conditional default values which is supported by other validators (e.g. in [Schema](https://github.com/keleshev/schema) using `Or` ). Here pledge for a new feature that adds the possibility to define more complex defaults i.g. where the default value depends on another field. This feature can be added without any new rule simply by considering `oneof`during normalization.

----

####  Feature request

It would be nice to consider `oneof` also for normalization. While other of-rules (`anyof`, `noneof`, `alloff`) are not unambiguous in the context of normalization `oneof` instead could be used to insert conditional default values.

e.g.
```
    schema = {
        'foo': {'type': 'string'},
        'bar' :{'oneof':[{'dependencies':{"foo":"B"}, 'default':"B"}, {'dependencies':{"foo":"A"}, 'default':"C"}]}
    }
    document = {'foo': 'A'}
    expected_normalization_result = {'foo': 'A','bar':"C"}
```

funkyfuture:
while you are correct that conditional normalization with the `oneof`-rule would seem to make sense, i still object this approach because of the inconsistency and additional complexity it introduces.

since you want to set a default value that depends on the document context, i propose to use a `default_setter`  with the extension proposed in #279. i have just started a branch where that is supposed to be implemented.

linupi:
thank you for your fast reply @funkyfuture and sorry that it took me so long to react.

At first I experimented with `default_setter` as well, but I didn't like it too much because it always sets a default, so it is not possible to insert a default only under certain conditions. 

On the validation side I really like the extensibility of Cerberus using ` _validate_MyNewMethod`, however I was lacking the equivalent for the normalization. That is how I started to think about the `oneof` for normalization, in order to avoid the need of extending Cerberus rules.

Following what is proposed in #279 I agree that the complexity inside the code might be reduced, but I am afraid that it makes the definition of the schema more complex and as far as I can see it might introduces redundancy in the schema definition as well when use for the problem that I tried to address here. However, I'd be happy to have a look once the corresponding branch is online - maybe my first impression is wrong.

[...]

PR #516, opened 2019-11-18
conditional normalization using oneof 

linupi:
I am not sure if I found the good entry-point to implement #515 but it certainly illustrates a possible path that could be taken to allow for  conditional normalization.

An open todo is in the schema of `_validate_oneof` where "'logical': 'oneof'" can no longer be used as it doesn't allow the `default` rule.


funkyfuture:
i'm closing this as #515.

PR #528, opened 2020-02-16
Fixed Typo in `conftest.py`.

bhanutejags:
Fixed Typo in `conftest.py`.


funkyfuture:
thanks for your attention. but it's not a typo, it's slang. i do not agree w/ the etymological section here, but :https://en.wiktionary.org/wiki/can_haz

PR #531, opened 2020-03-06
Adds a benchmark to compare overall performance for the 1.3 series

funkyfuture:

PR #532, opened 2020-03-06
Adds a benchmark to compare overall performance for the 2.x series

funkyfuture:
beside the adjusted benchmark from #531 this includes a fix for a revealed bug and minor changes.

the benchmark already proved useful as it showed the current `master` branch takes about three times of the `1.3.x` branch. i tracked down the impacting change commits, and while some are planned subjects for improvements and optimization, there are changes where i don't see why they would slow down the process like bf67dd9 and 37dd532 (i'll double check later).


funkyfuture:
merged as 8fa6c476ad9a9db650630c8bc52f6830f7d1a6bc, sorry the merge commit got lost  by squashing in a formatting issue.

PR #537, opened 2020-05-01
Fix missing schema error typo

stroykova:
Hello! 

Found this non-existing error code. 

I was not able to create a unit test for this. Can not guess a small example to reach this line. 
Is this line reachable? If there is any small example, I will add a unit test in this PR.

### DIFF
--- a/cerberus/schema.py
+++ b/cerberus/schema.py
@@ -271,7 +271,7 @@ class ValidatedSchema(MutableMapping):
             schema = self.validator.schema_registry.get(schema, schema)
 
         if schema is None:
-            raise SchemaError(errors.SCHEMA_MISSING)
+            raise SchemaError(errors.MISSING_SCHEMA)
 
         resolved = {
             k: self.validator.rules_set_registry.get(v, v)


funkyfuture:
> Can not guess a small example to reach this line. Is this line reachable?

that's weird, [here](https://github.com/pyeve/cerberus/blob/master/cerberus/tests/test_schema.py#L15) is the test covering it. it's also failing when one is providing an empty dictionary instead of `None`. maybe you could check out what happens when also the expected message is checked. see the next test for an example.

stroykova:
[test_empty_schema](https://github.com/pyeve/cerberus/blob/master/cerberus/tests/test_schema.py#L15) exits [earlier](https://github.com/pyeve/cerberus/blob/master/cerberus/base.py#L960) in __init_processing method. 

I assume [this](https://github.com/pyeve/cerberus/blob/master/cerberus/schema.py#L273) schema None check is about previous code block:

```
        if isinstance(schema, str):
            schema = self.validator.schema_registry.get(schema, schema)
```

So we need a test case when schema is string type and not in schema_registry. I tried [test_bad_schema_type](https://github.com/pyeve/cerberus/blob/master/cerberus/tests/test_schema.py#L21). But it also will exit [earlier](https://github.com/pyeve/cerberus/blob/master/cerberus/schema.py#L206). That is why I am not sure this code is reachable. Maybe there are some initialization steps or other more complicated usages.


About travis failure. I run ./run-docker-tests file multiple times in **master** branch and see non-deterministic behavior. All errors are in py35 env.

First attempt: 
```
>       assert _set['a_dict']['valuesrules']['regex'].errors[0] == _ref_err
E       assert ValidationError @ 0x7f5d02c11550 ( document_path=('a_dict', 'three'),schema_path=('a_dict', 'valuesrules', 'regex'),code=0x41,constraint="[a-z]*",value="abC",info=() ) == ValidationError @ 0x7f5d02c113c8 ( document_path=('a_dict', 2),schema_path=('a_dict', 'valuesrules', 'regex'),code=0x41,constraint="[a-z]*$",value="aBc",info=() )
```

Then after some successful attempts:
```
>           validator.schema = {'foo': {'type': 'unknown'}}
E           AssertionError: Pattern '\\{\\\'foo\\\'\\:\\ \\[\\{\\\'type\\\'\\:\\ \\[\\{0\\:\\ \\[\\\'none\\ or\\ more\\ than\\ one\\ rule\\ validate\\\'\\,\\ \\{\\\'oneof\\ definition\\ 1\\\'\\:\\ \\[\\"must\\ be\\ one\\ of\\ these\\ types\\:\\ \\(\\\'type\\\'\\,\\ \\\'generic_type_alias\\\'\\)\\"\\]\\,\\ \\\'oneof\\ definition\\ 0\\\'\\:\\ \\[\\\'Unsupported\\ type\\ name\\:\\ unknown\\\'\\]\\}\\]\\}\\]\\}\\]\\}' does not match '{\'foo\': [{\'type\': [{0: [\'none or more than one rule validate\', {\'oneof definition 0\': [\'Unsupported type name: unknown\'], \'oneof definition 1\': ["must be one of these types: (\'type\', \'generic_type_alias\')"]}]}]}]}'
```

I need about 5-10 attempts to see an error.
Should I open issues about that?

funkyfuture:
a) schema as string

i actually have newer code under my pillow where this is addressed. i just hadn't had the time yet to add the required docs. with these i'll open a pr.

b) the non-deterministic tests

yes, this is annoying for quiet a while now. it's caused by different sorting behaviours, interestingly i could never reproduce this locally. i once thought that it would go away when older interpreters are dropped. i think this is a test i want to refactor anyway, because there are more reasons why it is rather terrible. but again, that'd be a thing that should be included with my other changes. if you are very keen on tackling this (it's certainly no fun task), i could share my current state of affairs.

stroykova:
a) ok, close request?
b) I'm not sure I will be able to solve it. maybe refactor it and if not solved create an issue and share with community, maybe someone will help with that

funkyfuture:
> a) ok, close request?

yes. if you want to fix that for the `1.3.x` branch, create a new one.

funkyfuture:
thanks for reporting, i found that assertion to be checked in `__init_processing` already, so yes, this was dead code.

funkyfuture:
thanks for reporting, i found that assertion to be checked in `__init_processing` already, so yes, this was dead code.

PR #538, opened 2020-05-01
Fix documentation warnings and build instruction

stroykova:
Hello!

I followed all steps from the contribution guide and found there is no requirements-docs.txt file in docs folder. Also I propose fixes for documentation warnings and empty link on * sign.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ _build
 # Jetbrains
 .idea
 .vscode/
+.devcontainer/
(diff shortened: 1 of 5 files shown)


stroykova:
How do you build docs? Do you run `python setup.py install` or modify conf.py in docs folder to append project to sys.path? imho it would be good to add some instructions about that or commit conf.py maybe.

funkyfuture:
> How do you build docs?

that is mentioned just below your change in `contribute.rst`.

PR #559, opened 2020-12-10
Fix error handling in nested schemas

jsnjack:
Errors are not handled correctly when using nested schemas (see test_errors_in_nested_schemas). Cerberus raises exception when trying to get the field from schema. `self.schema` needs to be a Schema instance, not a string

### DIFF
--- a/cerberus/base.py
+++ b/cerberus/base.py
@@ -655,7 +655,8 @@ class UnconcernedValidator(metaclass=ValidatorMeta):
             if not rule:
                 constraint = None
             else:
-                field_definitions = self._resolve_rules_set(self.schema[field])
+                schema = self._resolve_schema(self.schema)
+                field_definitions = self._resolve_rules_set(schema[field])
                 if rule == 'nullable':
                     constraint = field_definitions.get(rule, False)
                 elif rule == 'required':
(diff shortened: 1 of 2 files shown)


funkyfuture:
ciao, thanks for contribution.

am i grasping correctly that the addressed bug only occurs when a schema registry is used? is the custom coercer needed to reproduce the bug?

did you notice the bug when using a 1.3 release or when working with the state from the master branch?

jsnjack:
Hi @funkyfuture,

Yeah, in order to reproduce this bug a schema registry with the custom coercer is needed. I've noticed this bug in 1.3 (with python 2.7), but it exists in both master and 1.3. Shall I create also (instead) a pr for 1.3 (this is the version we use at the moment) and if so, will it be released?

Thanks!

jsnjack:
Hi @funkyfuture,

Is there something else I need to do in order to get this pr merged? Please let me know :)

funkyfuture:
ciao @jsnjack, it's my lack of time. in particular reg. the `master` branch, b/c i have a lot of unpushed changes locally. anyway, yes, if you prepared am additional pr against the `1.3.x` branch that would help a lot. can't say anything about an estimated release though.

jsnjack:
thanks @funkyfuture ! I'll prepare the pr shortly

funkyfuture:
notes to self: the bug still exists in my local `annotated_handlers` branch, but the `_resolve_schema` method isn't available anymore.

funkyfuture:
closing this as we don't intend to produce a new major release and the fix is already included in the 1.3 series.

PR #566, opened 2021-04-08
Clarifies the regex rule's behaviour in the docs

funkyfuture:
@StevenMMortimer , it'd be great if could have a look at this. also note that i corrected my answer on StackOverflow.


funkyfuture:
solved with 2857c0e21e6fc9b9e15f4638589495567afa9dfa

Issue #568, opened 2021-05-17
v1.3.3 and v1.3.4 is not compatible with py38/py39


ssbarnea:
Despite the fact that `Cerberus==1.3.4` list compatibility with py39 we can easily get:
```
.tox/py39/lib/python3.9/site-packages/cerberus/validator.py:19: in <module>
    from cerberus import errors
E     File "/Users/ssbarnea/c/a/molecule/.tox/py39/lib/python3.9/site-packages/cerberus/errors.py", line 156
E       """
E       ^
E   SyntaxError: invalid escape sequence \*
```

While this bug was fixed in master it is not of any help for consumers of the library as there is no newer release. Even a pre-release like `2.0.0a0` could help as it would allow us to request it when running on py39 in setup.cfg.

Another approach would be backporting the fixed and making another hotfix release.

funkyfuture:
the error message isn't telling me anything. might it be related to an optimization level of the interpreter?

because [this runs fine](https://github.com/pyeve/cerberus/runs/2512883806) and i have an application here that runs on 3.9 with Cerberus 1.3.4.

tadeboro:
I am not sure if this has anything to do with Python 3.9. The docstring that the error points to contains an invalid escape sequence (`\*`) and that error is not related to python 3.9 at all.

It looks like https://github.com/pyeve/cerberus/commit/9844cb18ba62952e630cf2151b5bbceffe4b6c4e is a bad backport of https://github.com/pyeve/cerberus/pull/538 (the original PR uses `r""" ... """` while the backport does not add the `r` prefix).

This is what `flake8` has to say about this:

```
$ flake8 venv/lib/python3.9/site-packages/cerberus/errors.py
venv/lib/python3.9/site-packages/cerberus/errors.py:157: [W605] invalid escape sequence '\*'
venv/lib/python3.9/site-packages/cerberus/errors.py:185: [W605] invalid escape sequence '\*'
```

funkyfuture:
yes, this seems more likely to be the cause. but i'm still clueless under which circumstances the `\*` in docstings causes errors.

tadeboro:
I cannot replicate the error on my Linux machines, but it looks like the original error was produced on a mac, so maybe that makes the difference? But in any case, the Cerberus has a syntax error in it that should probably be fixed.

ssbarnea:
I should mention that I did not call flake8, that was a runtime error with molecule I seen on mac. Any of the two newer releases gave this error but the previous one worked.

[...]

PR #570, opened 2021-05-19
Backport docstring invalid syntax fixes

ssbarnea:
Backport these fixed in order to avoid runtime errors with newer python versions.

Fixes: #568

### DIFF
--- a/cerberus/errors.py
+++ b/cerberus/errors.py
@@ -154,7 +154,7 @@ class ValidationError(object):
     @property
     def definitions_errors(self):
         """
-        Dictionary with errors of an \*of-rule mapped to the index of the definition it
+        Dictionary with errors of an *of-rule mapped to the index of the definition it
         occurred in. Returns :obj:`None` if not applicable.
         """
         if not self.is_logic_error:
@@ -182,7 +182,7 @@ class ValidationError(object):
     @property
     def is_logic_error(self):
         """
-        ``True`` for validation errors against different schemas with \*of-rules.
+        ``True`` for validation errors against different schemas with *of-rules.
         """
         return bool(self.code & LOGICAL.code - ERROR_GROUP.code)
 
--- a/tox.ini
+++ b/tox.ini
@@ -25,4 +25,4 @@ commands=pre-commit run --config .linting-config.yaml --all-files
 
 [flake8]
 max-line-length=88
-ignore=E203,W503,W605
+ignore=E203,W503


ssbarnea:
@funkyfuture Please let it run and check if you can release a hotfix. This should buy us enough time until 2.0 is released.

funkyfuture:
regarding the failed test: this is due to an incompatibility of pytest via attrs with Python 3.4. i'll have to do some research before figuring out the best solution.

could you verify that this solved the issue with your problematic interpreter?

did you render the docs to html and have a look at the result where the escaped asterisk is used in the docstrings?

ssbarnea:
Yes it does fix the issue, that is why I raise the PR.

The docs compiles fine and renders correctly. Not sure why the project is not using rtd builds, which now can also build PRs and allow user to navigate the built side from the PR (very useful for reviewing changes that impact the docs).

BTW, I would not call it problematic interpreter, the python spec deprecated invalid string escapes in 3.6 and mentioned they will become errors in the future. If I remember correctly I got the same error with py38, but not with older versions so I suspect these newer versions are pickier. 

IMHO, `W605` should have never being ignored in the first place.

funkyfuture:
> Yes it does fix the issue, that is why I raise the PR.

since the bug isn't reproducible on the CI and you didn't mention it, i had to ask.

>Not sure why the project is not using rtd builds, which now can also build PRs and allow user to navigate the built side from the PR (very useful for reviewing changes that impact the docs).

the docs are built on rtd. the pull requests feature is new to me and i'm not sure whether it's worth the effort. if you think that it'd be a helpful addition, please open an issue.

> I would not call it problematic interpreter

excuse my sharp language. i wanted to stress out that you are the only user reporting this issue so far and that we couldn't reproduce it elsewhere yet. however, a reproducible error would ensure that something similar will not be re-introduced. i think that's important.

> the python spec deprecated invalid string escapes in 3.6

could you please point me to a resource where this is mentioned, as i can't find it [in Python's documentation](https://docs.python.org/3.6/whatsnew/3.6.html) (where i searched for instances of "string").

funkyfuture:
quick note: i verified the resulting API docs and the unescaped `*` token leads to a dead-ended anchor link,

ssbarnea:
That is a different issue, likely a double backslash is needed. Using special chars in a link is never a good idea anyway (browser will need to encode them)

hswong3i:
I merge this PR as hack for my OBS packaging:
* https://github.com/alvistack/pyeve-cerberus/commit/34767813c9d804238e5a3eaa0da6ae6777f4a140
* https://build.opensuse.org/package/show/home:alvistack/pyeve-cerberus-1.3.4

When combine with a manual hack for molecule OBS packaging:
* https://github.com/alvistack/ansible-community-molecule/commit/d5d36c33ff6f38cb08e2d6ede7cb5f34a682f799
* https://build.opensuse.org/package/show/home:alvistack/ansible-community-molecule-3.5.2

They are now working fine all together with my Ubuntu 21.10 + Python 3.9.7. Thank you very much.

ssbarnea:
Cutting the loses...

PR #579, opened 2021-10-05
decrease memory consumption, use importlib.metadata to get package version instead of pkg_resources.get_dictribution

akayunov:
pkg_resources use a lot of memory and after introducing importlib.metadata in 3.8 looks obsolete. So let's change way to get package version and get rid of pkg_resources import.

### DIFF
--- a/cerberus/__init__.py
+++ b/cerberus/__init__.py
@@ -8,7 +8,6 @@
 
 """
 
-import sys
 from typing import Dict, Optional, Tuple, Union
 
 from cerberus.base import (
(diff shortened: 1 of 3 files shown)


akayunov:
Do we really need tests like these: cerberus/tests/test_assorted.py:23?

funkyfuture:
thanks for bringing this up. i think this would be well covered as the test suite runs on all supported platforms.

could you please check whether the `importlib_metadata` object could be defined in and imported from the `platform` module?

you may have noticed that development has stalled for a while and it will take such to merge and release this pr.

funkyfuture:
oh, iirc this test is covering some edge case. please try to do some research in the commit / pr / issues history.

slanzmich:
Is there any chance to get this merged and released? If yes, I can provide a fix for the tests in cerberus/tests/test_assorted.py.

setup.py
```diff
@@ -40,7 +40,10 @@
     setup_requires=setup_requires,
     tests_require=["pytest"],
     test_suite="cerberus.tests",
-    install_requires=["setuptools"],
+    install_requires=[
+        "setuptools",
```
socketpair, line 44: NO! please remove setuptools. It's not required actually!

evelikov:
@funkyfuture sorry to bother, but can we get this merged (or variation thereof) and a new release? Thanks in advance

cerberus/__init__.py
```diff
@@ -21,11 +20,9 @@
 from cerberus.schema import SchemaError
 from cerberus.validator import Validator
 
+from cerberus.platform import importlib_metadata
 
-try:
-    __version__ = get_distribution("Cerberus").version
-except DistributionNotFound:
-    __version__ = "unknown"
+__version__ = importlib_metadata.version("Cerberus")
```
socketpair, line 25: I suggest just dropping it. It's not required.

funkyfuture:
solved w/ 81e2dfd6f79847e68133ce42b78dc3bb55870268

PR #580, opened 2021-12-04
Note about Validator thread-safety

pykler:
the `validate*` methods of a validator makes a copy of the document being
validated and rely on that document not changing during validation.
When used in a multi-threaded context with a global validator object 
(such as in a webserver or a REST api), it is possible for the document to change
while the validation is taking place which causes seemingly random errors.
This note helps users understand that validators are not thread safe.


funkyfuture:
thanks, this is included w/ 0ffe3a3c840e480f04d647fd97a0d0dae9f7da78 & 4ae57ef31f252a4223cab244a71928f575d40326

Issue #628, opened 2025-11-25
`lidationError.__lt__` returns `True` for identical operands


T90REAL:
`ValidationError.__lt__` relies on `cerberus.utils.compare_paths_lt`. That helper currently returns `True` whenever the common prefix of two paths matches their entire length. When both paths are identical tuples, the prefix check succeeds and the function still returns `True`, so `a < a` evaluates to `True`, which actually violates Python’s strict ordering contract.

```python
from cerberus.errors import ValidationError
from cerberus.utils import compare_paths_lt

err = ValidationError((), (), 0, '', None, None, ())
print(compare_paths_lt(err.document_path, err.document_path))
print(err < err)
```

```
True
True
```

funkyfuture:
we never signed that strict ordering contract as in the implementation is complete enough for its purpose. but you can provide a fix if you feel the need.

PR #630, opened 2025-11-25
fix `compare_paths_lt` irreflexivity for equal paths

T90REAL:
Solve #628.
Fix `compare_paths_lt` to align with Python’s strict ordering contract.

### DIFF
--- a/cerberus/utils.py
+++ b/cerberus/utils.py
@@ -21,7 +21,7 @@ def compare_paths_lt(x, y):
     min_length = min(len(x), len(y))
 
     if x[:min_length] == y[:min_length]:
-        return len(x) == min_length
+        return len(x) < len(y)
 
     for i in range(min_length):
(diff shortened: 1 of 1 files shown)


funkyfuture:
please provide tests that have a practical impact where you prove that the current implementation isn't sufficient.

if you propose such a significant change, also add tests that cover the possible edge cases between the current state and your proposal, so we can be sure that the current sorting behaviour persists.

then (again), i don't know where you take that strong contract from and why you think it would be mandatory in that case.

T90REAL:
Thanks for your response. My original thought is that if `err < err == True` would cause some function does not work as designed, for example the `bisect_left` function. After reconsidering the context, I agree it doesn’t have a significant impact on the behavior. I’ll close the PR here, sorry for the noise.

funkyfuture:
well, let's consider this an academic exercise.

PR #642, opened 2026-07-31
Remove `importlib_metadata.version` usage 

### LINKED ISSUE #640: Remove `importlib_metadata.version` usage
It is not used anywhere except exporting `__version__` of Cerberus itself.  Since it double underscore property - it IS private. so no  one should be using it. If one need to get version, he can get it using 

`importlib_metadata.version("Cerberus")`

In the past, I tried: #603, but removal was declined in fovour of code refactoring. I insist on removing. List of places to remove:

https://github.com/search?q=repo%3Apyeve%2Fcerberus+importlib_metadata&type=code

If you agree - I can make PR.

socketpair:
Fix #640


socketpair:
@funkyfuture please review

.github/workflows/tests.yml
```diff
@@ -29,7 +29,6 @@ jobs:
           - 3.12  # 2028-10
           - 3.13  # 2029-10
           - 3.14  # 2030-10
-          - pypy3.7
```
socketpair, line 32: Not supported after this PR. Python 3.7 EOL Date: June 27, 2023

socketpair:
@funkyfuture

funkyfuture:
as mentioned in some issue, we're not breaking support for older versions, even if they're eol as we will only release service releases.

PR #647, opened 2026-09-17
Fix mutable default values being shared across normalized documents

afonsojanu:
`Validator.normalized()` copies a schema's `default` value straight onto the output document:

```python
def _normalize_default(self, mapping, schema, field):
    """{'nullable': True}"""
    mapping[field] = schema[field]['default']
```

When the default is immutable (a string, a number) that's fine, but when it's a list, dict, or any other mutable object, every document normalized against that schema ends up holding a reference to the exact same object. A schema is usually defined once and reused for many documents, so this bites as soon as you mutate a default-populated field on one document:

```python
from cerberus import Validator

schema = {'tags': {'type': 'list', 'default': []}}
v = Validator(schema)

doc1 = v.normalized({})
doc1['tags'].append('a')

doc2 = v.normalized({})
print(doc2['tags'])  # ['a'] -- leaked from doc1, should be []
print(doc1['tags'] is doc2['tags'])  # True
```

This isn't limited to reusing one `Validator` instance either. Since the shared object lives on the schema dict itself, two completely separate `Validator(schema)` calls against the same schema definition exhibit the same leak.

The fix deep-copies the default before assigning it, so each normalized document gets an independent value. Added a regression test that reproduces the leak with a list default, and bumped `max_cache_size` in `test_validated_schema_cache` by one, as its own docstring/comment asks contributors to do when a new test introduces a new schema.


funkyfuture:
you can use the `default_setter` rule for such cases.

